PR #4, opened 2024-04-08
[BSENG-2318] Switch to GitHub runners, use matrix strategy for test execution

rgildein:
The small self-hosted runners were removed and since we do not need to run this workflow on self-hosted runner, I'll change to GitHub runners. 

I also change to use matrix strategy with [max-parallel](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#defining-the-maximum-number-of-concurrent-jobs) parameter set to 3 and adding also documentation about it. However this required to use artifacts to store results, so I update also that with new message.

run example: [23668147146](https://github.com/canonical/solutions-engineering-automation/actions/runs/8633904492/job/23668147146)
message example:
Screenshot from 2024-04-10 17-42-34

### DIFF
--- a/.github/workflows/weekly_tests.yaml
+++ b/.github/workflows/weekly_tests.yaml
@@ -10,8 +10,6 @@ jobs:
     name: Run weekly tests
     runs-on: ubuntu-latest
     timeout-minutes: 1440
-    outputs:
-      failed_tests: ${{ steps.collect-failed-tests.outputs.failed_tests }}
     strategy:
       max-parallel: 3
       fail-fast: false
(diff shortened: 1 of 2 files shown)


chanchiwai-ray (approved):
LGTM, thanks! (The MM notification looks nice now)

sudeephb (approved):
LGTM

.github/workflows/weekly_tests.yaml
```diff
@@ -6,410 +6,142 @@ on:
...
+                  "job-index": "${{ strategy.job-index }}",
+                  "repo": "${{ matrix.repo }}",
+                  "branch": "${{ matrix.branch }}",
+                  "workflow_file_name":"${{ matrix.workflow_file_name }}",
+                  "conclusion": "${{ steps.dispatched-tests.outcome }}",
+                  "workflow_url": "${{ steps.dispatched-tests.outputs.workflow_url }}"
+              }
+          END
+          )
+          echo $result > "result-${{ strategy.job-index }}.json"
```
agileshaw, line 118: non-blocking: It would be better to use repo+workflow name instead of `strategy.job-index` so a human can identify a file quickly when troubleshooting

.github/workflows/weekly_tests.yaml
```diff
@@ -6,410 +6,142 @@ on:
...
+
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: "result-${{ strategy.job-index }}"
+          path: "result-${{ strategy.job-index }}.json"
+
+  notify:
+    needs: weekly-tests
+    if: always()  # these needs to be run always
```
agileshaw, line 128:
```suggestion
    if: always()  # these need to be run always
```

agileshaw (commented):
Mostly LGTM. Just one typo and a non-blocking suggestion

.github/workflows/weekly_tests.yaml
rgildein: We can use all parameter from matrix, but these artifacts are only for internal usage, and you should found the same information in each job, so I do not see reason for anyone to go there.

agileshaw: I still think it would be better having a more human-readable filename, but this is not at all a blocker. Resolving.

agileshaw (approved):
LGTM, thanks :)

chanchiwai-ray (approved):

sudeephb (approved):

Pjack (approved):

README.md
```diff
@@ -1,2 +1,18 @@
...
+
+## How it works
+
+The weekly_tests workflow lists all repos on GitHub maintained by the Solutions Engineering Team, and the job for dispatching test on top of those repos remotely.
+The configuration description:
+ - `repo` is the name of a repository in canonical organization, e.g. charm-apt-mirror
+ - `workflow_file_name` is name of workflow file which should be triggered, e.g. check.yaml
+ - `branch` name of branch from which tests should be executed, e.g. main
+
+Jobs are running in parallel with the maximum number of concurrency defined by the `max-parallel` strategy parameter,
```
dashmage, line 13: nit: Would be nice to add a link to the docs page under "max-parallel" with https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#defining-the-maximum-number-of-concurrent-jobs

dashmage (approved):
LGTM :rocket: Just left one small comment which isn't a big deal even if it isn't addressed.

Thanks for these changes. The workflow looks much cleaner and easier for adding a new repo after refactoring the steps into a matrix. And nice usage of artifacts to collect all the results to craft the MM message in the later job. The final message is also much more informative providing outcomes for both successful and failed runs.

dashmage:
Another minor nit: Could we also change the PR title to something like: "Switch to GitHub runners, use matrix strategy for test execution" to be more relevant to the changes?

rgildein:
> Another minor nit: Could we also change the PR title to something like: "Switch to GitHub runners, use matrix strategy for test execution" to be more relevant to the changes?

This starts as simple switch to GitHub runners, and since it require some fixes also I change much more. Changed thanks

README.md
rgildein: Created #8.

PR #14, opened 2024-05-17
first POC with GH settings and CODEOWNERS

rgildein:
This is an enhancement for trying to GitHub repo settings via terraform and providing also files like CODEOWNER.

### DIFF
--- a/.github/workflows/terraform-apply.yaml
+++ b/.github/workflows/terraform-apply.yaml
@@ -1,0 +1,51 @@
+name: Apply new changes to SolEng repositories
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - terraform-plans/**
+  on:
+    push:
+      branches:
+        - main
+      paths:
+        - terraform-plans/**
+
+jobs:
+  runs-on: ubuntu-latest
+  env:
+    GITHUB_APP_ID: ${{ secrets.SOLENG_APP_ID }}
+    GITHUB_APP_INSTALLATION_ID: ${{ secrets.SOLENG_APP_INSTALLATION_ID }}
+    GITHUB_APP_PEM_FILE: ${{ secrets.SOLENG_APP_PEM_FILE }}
+  strategy:
+      fail-fast: false
+      matrix:
+        repository:
+          - soleng-tf-test-repo
+  steps:
+    - name: Checkout branch
+      uses: actions/checkout@v4
+    - name: Install Terraform
+      uses: sudo snap install terraform
+    - name: Terraform init
+      working-directory: ./terraform-plans
+      run: terraform init
+    - name: Terraform validate
+      working-directory: ./terraform-plans
+      run: terraform validate --no-color
+    - name: Terraform plan
+      working-directory: ./terraform-plans
+      run: |
+        terraform plan -no-color \
+        -var-file=configs/github.tfvars \
+        -var-file=configs/${{ matrix.repository }}.tfvars
+    - name: Terraform Apply only on merge to main
+      if: ${{ github.event_name == 'push' }}
+      working-directory: ./terraform-plans
+      run: |
+        terraform apply -no-color -auto-approve \
+        -var-file=configs/github.tfvars \
+        -var-file=configs/${{ matrix.repository }}.tfvars
--- a/.github/workflows/terraform-apply.yaml
+++ b/.github/workflows/terraform-apply.yaml
@@ -7,45 +7,54 @@ on:
       - main
     paths:
       - terraform-plans/**
-  on:
-    push:
-      branches:
-        - main
-      paths:
-        - terraform-plans/**
+  push:
+    branches:
+      - main
+    paths:
+      - terraform-plans/**
 
 jobs:
-  runs-on: ubuntu-latest
-  env:
-    GITHUB_APP_ID: ${{ secrets.SOLENG_APP_ID }}
-    GITHUB_APP_INSTALLATION_ID: ${{ secrets.SOLENG_APP_INSTALLATION_ID }}
-    GITHUB_APP_PEM_FILE: ${{ secrets.SOLENG_APP_PEM_FILE }}
-  strategy:
-      fail-fast: false
-      matrix:
-        repository:
-          - soleng-tf-test-repo
-  steps:
-    - name: Checkout branch
-      uses: actions/checkout@v4
-    - name: Install Terraform
-      uses: sudo snap install terraform
-    - name: Terraform init
-      working-directory: ./terraform-plans
-      run: terraform init
-    - name: Terraform validate
-      working-directory: ./terraform-plans
-      run: terraform validate --no-color
-    - name: Terraform plan
-      working-directory: ./terraform-plans
-      run: |
-        terraform plan -no-color \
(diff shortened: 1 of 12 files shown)


terraform-plans/configs/github.tfvars
```diff
@@ -0,0 +1 @@
+owner = "Canonical"
```
chanchiwai-ray, line 1: Compliant by CI: Maybe "canonical" with a lower case 'c'?

rgildein: I tried that localy, same. The app need to have read-only for organization/members.

.github/workflows/terraform-apply.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+jobs:
+  terraform:
+    name: Run Terraform
+    runs-on: ubuntu-latest
+    strategy:
+        fail-fast: false
+        matrix:
+          repository:
+            - soleng-tf-test-repo
```
chanchiwai-ray, line 24: (non-blocking) This can be embedded in `tfvars` and using `for_each`, but there are pros and cons to it

terraform-plans/modules/GitHub/settings/main.tf
```diff
@@ -0,0 +1,87 @@
...
+  has_wiki        = false
+  has_discussions = false
+
+  allow_merge_commit = false
+  allow_squash_merge = true
+  allow_rebase_merge = false
+  allow_auto_merge   = false
+
+  allow_update_branch    = true
+  delete_branch_on_merge = true
```
Pjack, line 29: Did we discuss this one during Madrid engineering sprint? If not, please set it as false (default value in github).

terraform-plans/modules/GitHub/settings/main.tf
```diff
@@ -0,0 +1,87 @@
...
+  team {
+    permission = "push"
+    team_id    = data.github_team.engineering.id
+  }
+}
+
+resource "github_branch_protection" "branch_protection" {
+  repository_id                   = github_repository.repo.node_id
+  pattern                         = var.branch
+  enforce_admins                  = true
```
Pjack, line 67: What is enforce_admin?

terraform-plans/modules/GitHub/settings/main.tf
```diff
@@ -0,0 +1,87 @@
...
+  require_signed_commits          = true
+  required_linear_history         = true
+  require_conversation_resolution = true
+  allows_deletions                = false
+
+  allows_force_pushes  = true
```
Pjack, line 73: Should this configuration (allow_force_push) to be false?

Pjack (changes requested):

terraform-plans/modules/GitHub/settings/main.tf
jneo8: I am fine with this one. No reason to keep the merged branch.

terraform-plans/modules/GitHub/settings/main.tf
jneo8: Can be found here: https://registry.terraform.io/providers/integrations/github/latest/docs/resources/branch_protection

> (Optional) Boolean, setting this to true enforces status checks for repository administrators.

terraform-plans/README.md
```diff
@@ -0,0 +1,48 @@
+# Terraform plans for Solution Engineering
+
+Currently these plans are only for setting GitHub repos and to add workflow files.
+
+## How to start
```
jneo8, line 5: **suggestion (non-blocking):** 

The document here doesn't look like what we need.

I expect some document to explain the whole idea of using terraform and basic idea of how the workflow will be triggered and maintained.

.github/workflows/terraform-apply.yaml
rgildein: I was not working correctly for my using tfvars, but we can try it again, maybe I did something wrong.

terraform-plans/modules/GitHub/settings/main.tf
rgildein: I'm ok with both, people often forgot to delete a branch, and then we have the lot of not relevant branches.

terraform-plans/modules/GitHub/settings/main.tf
rgildein: To be honest, I have no idea what it's actually doing, but I saw that in all examples.

terraform-plans/README.md
rgildein: yes, I did not provide much since this was just POC. I'll update the document after reviews.

rgildein: I do not think that I will have time to do this, so we can do it in separate PR or someone else can do it here.

jneo8: This one will handover to me. Open #16 to trace.

jneo8 (approved):
Open #16 and #17 to trace issues. Other is LGTM.

.github/workflows/terraform-apply.yaml
chanchiwai-ray: it depends on which var you want to iterate over; you choose to iterate over the template files, and provide repository one-by-one. Alternatively, you can choose to iterate over repositories, and define template files resource one-by-one. The latter provide a better `graph` for visualizing what will be added e.g. you can do `terraform graph | dot -Tsvg > output.svg`

terraform-plans/modules/GitHub/settings/main.tf
chanchiwai-ray: Maybe leave it as an option (i.e. `false`); enabling it will block people from trying different development pattern (e.g. working on a dev branch for several small features, etc. No that we will do very often though)

Pjack: Please set it as "false" in this PR. We can change it after the discussion in the team.

rgildein: Done.

.github/workflows/terraform-apply.yaml
rgildein: I tried to iterate over repositories, however there was an issue with usage of modules. I was not able to import staff and modules did not have access to github provider (the app credentials). Fill free to open another PR and try it, I did not manage it before Madrid

jneo8 (approved):

.github/workflows/terraform-apply.yaml
chanchiwai-ray: I tried importing from root module (with my approach), and it worked fine. But I would rather not use import because it cannot address many issue (e.g. `override_on_create` is not imported correctly, pull request cannot be imported), so I would rather leave the import for github files, and have a workflow step to check if there are any update to the file. I am resolving it now, we can iterate more as we see needed

terraform-plans/modules/GitHub/settings/main.tf
```diff
@@ -0,0 +1,87 @@
...
+      version = "~> 6.0"
+    }
+  }
+}
+
+provider "github" {
+  owner = var.owner
+  app_auth {} # using environmet variables for auth
+}
+
```
chanchiwai-ray, line 14: (non-blocking) it should be inherented from root module, so probably useless

Pjack (approved):

PR #18, opened 2024-05-29
Doc/terraform readme

jneo8:
Follow: #14


terraform-plans/README.md
```diff
@@ -1,48 +1,21 @@
...
+> For more details how to use Github application authentication, please check [Terraform Github Provider - Github App Installation](https://registry.terraform.io/providers/integrations/github/latest/docs#github-app-installation)
+
+The permissions required for the github application are:
+
+-  Read access to members and metadata
+-  Read and write access to actions, actions variables, administration, checks, code, commit statuses, environments, issues, merge queues, packages, pull requests, secrets, and workflows
+
+> This Github Application need to be installed on every repositories we want to manage. Please ask people who has Github Organization permission for help.
+> 
+> For how to create the github application, please check [Creating Github Apps](https://docs.github.com/en/apps/creating-github-apps)
```
chanchiwai-ray, line 21: is this true? I think the bot should only need to have access to the required repo

rgildein: What do you mean by this?

jneo8: > is this true? I think the bot should only need to have access to the required repo

Yes, every repositories we want to manage == required repo

chanchiwai-ray: so every required repo need to install the bot, not just this repo?

jneo8: As I understand, yes. Because the permission flow is like:

`github_app_secrets(soleng-auto repo's secrets)` --> `CI on soleng-auto repo` --> `terraform` --> `all required repositories`

The permission to access those repositories is given by the github application, which be installed on the repositories by the github org admin.

rgildein: yes, that is true.

chanchiwai-ray: Okay thanks for the clarification. This seems one step forward and one step backward... (it's not so much different from having an action in each repository to sync with a central repository, except with a bot's secret)

rgildein: Action in repo is potentially good solution and action do not need to have any secrets to access central repo, since our repos are open-source. However, this action will need to have secrets or permission (I think you can set permission for workflow to create PRs).
The main draw back for me would be that you need to check each repo if the action works fine, instead of doing it from one place.

chanchiwai-ray (approved):

Pjack (approved):

PR #26, opened 2024-06-04
fix(terraform): Fix wrong repo configuration

jneo8:
- For branch protection rule
  - Remove force_push_bypassers
  - Remove dismissal_restrictions
- Add vulnerability_alerts=true for repo setting

### DIFF
--- a/terraform-plans/main.tf
+++ b/terraform-plans/main.tf
@@ -4,8 +4,6 @@ module "github_settings" {
   repository             = var.repository
   repository_description = var.repository_description
   branch                 = var.branch
-  force_push_bypassers   = ["${var.owner}/soleng-admin"]
-  dismissal_restrictions = ["${var.owner}/soleng-admin", "${var.owner}/soleng-reviewers"]
   pull_request_bypassers = ["${var.owner}/soleng-admin"]
 }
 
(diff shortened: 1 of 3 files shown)


Pjack (approved):
Thanks to adjust the configuration again based the [doc](https://docs.google.com/document/d/12dwn3lvewX4B5TJTjG1CwFpcr8-62Viy1o3B9Oe9Bhg/edit)

terraform-plans/main.tf
```diff
@@ -4,8 +4,6 @@ module "github_settings" {
   repository             = var.repository
   repository_description = var.repository_description
   branch                 = var.branch
-  force_push_bypassers   = ["${var.owner}/soleng-admin"]
-  dismissal_restrictions = ["${var.owner}/soleng-admin", "${var.owner}/soleng-reviewers"]
   pull_request_bypassers = ["${var.owner}/soleng-admin"]
```
Pjack, line 7: pull_request_bypassers 
This can be removed.

Pjack (changes requested):

Pjack (approved):

rgildein:
FYI: Those settings were there to mimic what was possible before. Terraform will set "Do not allow bypassing the above settings ", so without any bypass even admin of repo should not do it.

PR #29, opened 2024-06-06
ci: Dropdown for promote.yaml

jneo8:
- Provide default track and risk levels as dropdown in promote.yaml

(planned but not in scope for this PR) - For some projects need to release to different track/risk-levels, we provide another customized file instead using standard one.

### DIFF
--- a/terraform-plans/files/github/charm_promote.yaml
+++ b/terraform-plans/files/github/charm_promote.yaml
@@ -1,4 +1,4 @@
-name: Promote charm to other tracks and channels
+name: Promote charm to default track, standard risk levels.
 
 on:
   workflow_dispatch:
@@ -6,9 +6,17 @@ on:
       destination-channel:
         description: 'Destination Channel, e.g. latest/candidate'
         required: true
+        type: choice
+        options:
+          - 'latest/candidate'
+          - 'latest/stable'
       origin-channel:
         description: 'Origin Channel, e.g. latest/edge'
         required: true
+        type: choice
+        options:
+          - 'latest/edge'
+          - 'latest/candidate'
 
 jobs:
   promote-charm:
--- a/terraform-plans/files/github/charm_promote.yaml
+++ b/terraform-plans/files/github/charm_promote.yaml
@@ -3,20 +3,13 @@ name: Promote charm to default track, standard risk levels.
 on:
   workflow_dispatch:
     inputs:
-      destination-channel:
-        description: 'Destination Channel, e.g. latest/candidate'
+      channel-promotion:
+        description: 'Channel Promotion, e.g. latest/edge -> latest/candidate'
         required: true
         type: choice
         options:
-          - 'latest/candidate'
-          - 'latest/stable'
-      origin-channel:
-        description: 'Origin Channel, e.g. latest/edge'
-        required: true
-        type: choice
-        options:
-          - 'latest/edge'
-          - 'latest/candidate'
+          - 'latest/edge -> latest/candidate'
+          - 'latest/candidate -> latest/stable'
 
 jobs:
   promote-charm:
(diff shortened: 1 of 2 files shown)


terraform-plans/files/github/charm_promote.yaml
```diff
@@ -1,14 +1,22 @@
...
+        options:
+          - 'latest/candidate'
+          - 'latest/stable'
       origin-channel:
         description: 'Origin Channel, e.g. latest/edge'
         required: true
+        type: choice
+        options:
+          - 'latest/edge'
+          - 'latest/candidate'
```
rgildein, line 19: What do you think about using something like this:
```yaml
options:
          - 'latest/edge -> latest/candidate'
          - 'latest/candidate -> latest/stable'
```
I know that implementation would require some changes, but it will be more bullet proof for user.

rgildein (commented):
I think it's ok, just one question for potential improvement.

terraform-plans/files/github/charm_promote.yaml
jneo8: The implementation will be something like:


```
name: Promote charm to default track, standard risk levels.

on:
  workflow_dispatch:
    inputs:
      channel-promotion:
        description: 'Channel Promotion, e.g. latest/edge -> latest/candidate'
        required: true
        type: choice
        options:
          - 'latest/edge -> latest/candidate'
          - 'latest/candidate -> latest/stable'

jobs:
  promote-charm:
    name: Promote charm
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Set channels
        id: set-channels
        run: |
          if [[ "${{ github.event.inputs.channel-promotion }}" == 'latest/edge -> latest/candidate' ]]; then
            echo "destination-channel=latest/candidate" >> $GITHUB_ENV
            echo "origin-channel=latest/edge" >> $GITHUB_ENV
          elif [[ "${{ github.event.inputs.channel-promotion }}" == 'latest/candidate -> latest/stable' ]]; then
            echo "destination-channel=latest/stable" >> $GITHUB_ENV
            echo "origin-channel=latest/candidate" >> $GITHUB_ENV
          fi
      - name: Release charm to channel
        uses: canonical/charming-actions/release-charm@2.4.0
        with:
          credentials: ${{ secrets.CHARMHUB_TOKEN }}
          github-token: ${{ secrets.GITHUB_TOKEN }}
          destination-channel: ${{ env.destination-channel }}
          origin-channel: ${{ env.origin-channel }}
```

I haven't test it, how do you think @Pjack ?

Pjack: This is my preferred user behavior. If the code is maintainable, then I'm happy with this.

jneo8: > `origin, destination = channel-promotion.split('->').map(strip_whitespace)`

I don't think this is doable on workflow yaml.

jneo8:
> > For some projects need to release to different track/risk-levels, we provide another customized file instead using standard one.
> 
> This is mentioned in the description, but is not present in the patch?

Yes, I try to explain the policy we will have.

terraform-plans/files/github/charm_promote.yaml
jneo8: I put the implementation on latest commit. Please check it there.

terraform-plans/files/github/snap_promote.yaml
```diff
@@ -1,14 +1,22 @@
...
 on:
   workflow_dispatch:
     inputs:
       destination-channel:
         description: 'Destination Channel, e.g. latest/candidate'
         required: true
+        type: choice
+        options:
+          - 'latest/candidate'
+          - 'latest/stable'
```
rgildein, line 12: Can we use same option as we did for charm. Thanks

rgildein (changes requested):
Please address @samuelallan72

chanchiwai-ray (approved):

terraform-plans/files/github/snap_promote.yaml
jneo8: It's updated.

rgildein (approved):
Thanks :+1:

PR #31, opened 2024-06-13
Document source and update process for templates

samuelallan72:
This is so it's clear how these files are managed, and how to update them,
when working with them in the target repositories.


jneo8 (approved):
Nice! LGTM

chanchiwai-ray (approved):

Issue #7, opened 2024-04-11
Simplify job matrix configuration for better readability and maintainability 


agileshaw:
Right now (#4) in weekly-test job matrix strategy, we are using `include` to verbosely list all repositories and their parameter (workflow, branch) one by one. This is necessary because we haven't unify the name of their default branches and workflow to call. Once the unifying work is done, I suggest rewrite the matrix configuration the following way so that it is easy to spot: 1) the list of repositories we are running weekly tests for; 2) what the standard parameter combinations are; 2) the few "irregular" test setup.

```yaml
jobs:
  weekly-tests:
    ...
    strategy:
        max-parallel: 3
        fail-fast: false
        matrix:
          repo: [charm-apt-mirror, charm-apt-mirror, charm-cloudsupport, charm-duplicity, hardware-observer-operator, ...]
          workflow_file_name: [check.yaml]
          branch: [main]
          include:  # only the special ones are listed explicitly
             - repo: hardware-observer-operator
               workflow_file_name: cos_integration.yaml
               branch: master
             - ...
```

PR #34, opened 2024-07-30
Simplify weekly tests job matrix

samuelallan72:
- Remove branch variable because all are using branch main now.
- Most are using check.yaml for the workflow file, so use that as the default, and add include entries for outliers.

Fixes: #7

### DIFF
--- a/.github/workflows/weekly_tests.yaml
+++ b/.github/workflows/weekly_tests.yaml
@@ -14,94 +14,45 @@ jobs:
       max-parallel: 3
       fail-fast: false
       matrix:
+        repo:
+          - charm-advanced-routing
+          - charm-apt-mirror
+          - charm-cloudsupport
+          - charm-duplicity
+          - charm-juju-backup-all
+          - charm-juju-local
+          - charm-kubernetes-service-checks
+          - charm-local-users
+          - charm-logrotated
+          - charm-nginx
+          - charm-nrpe
+          - charm-openstack-service-checks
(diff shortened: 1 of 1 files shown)


Pjack (approved):

samuelallan72:
Note: untested but I guess we can compare with previous weekly tests to see if anything changed/broke.

.github/workflows/weekly_tests.yaml
```diff
@@ -14,94 +14,45 @@ jobs:
...
-            workflow_file_name: check.yaml
-            branch: main
-          - repo: charm-userdir-ldap
-            workflow_file_name: check.yaml
-            branch: main
-          - repo: charmed-openstack-upgrader
-            workflow_file_name: check.yaml
-            branch: main
           - repo: hardware-observer-operator
             workflow_file_name: check.yaml
```
chanchiwai-ray, line 47: did you forget to include this in `matrix`?

samuelallan72: No I put it here to be clearer, because it's there twice (once for check.yaml, once for cos_integration.yaml).

chanchiwai-ray: isn't both `check.yaml`?

chanchiwai-ray: oh, sorry. I must be blind...

jneo8 (approved):

PR #38, opened 2024-08-02
Refactor workflow_files variable

samuelallan72:
Rename workflow_files to templates:
This makes it clearer that these are any files that are centrally managed via templates from this repository. It's not just workflow files (precedent being CODEOWNERS file).

Change workflow_files.variables type and name:
- type `any` will be required if we want template variables of other types - for example a list of architectures to run on (useful for multiarch releases coming soon)
- vars name is more consistent with the templatefile function args
- also reduce a level of nesting (pass this directly to the template file function, and access the variables directly in the template instead of vars.name)

### DIFF
--- a/terraform-plans/main.tf
+++ b/terraform-plans/main.tf
@@ -6,34 +6,34 @@ module "github_settings" {
   branch                 = var.branch
 }
 
-module "github_workflow_files" {
-  source         = "./modules/GitHub/workflows"
+module "github_templates" {
+  source         = "./modules/GitHub/templates"
   owner          = var.owner
   repository     = var.repository
   branch         = var.branch
-  workflow_files = var.workflow_files
+  templates = var.templates
 }
 
 output "repository" {
-  value = module.github_workflow_files.repository
+  value = module.github_templates.repository
 }
 
 output "branch" {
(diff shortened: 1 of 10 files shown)


terraform-plans/main.tf
```diff
@@ -6,34 +6,34 @@ module "github_settings" {
...
 
-module "github_workflow_files" {
-  source         = "./modules/GitHub/workflows"
+module "github_templates" {
+  source         = "./modules/GitHub/templates"
   owner          = var.owner
   repository     = var.repository
   branch         = var.branch
-  workflow_files = var.workflow_files
+  templates = var.templates
```
Pjack, line 14: nitpick, different space with previous line, please align them if possible. Please solve the conversation by yourself after fix it.

Pjack (changes requested):
LGTM, only a small nitpick

terraform-plans/main.tf
samuelallan72: Thanks, done. :)

jneo8 (approved):

chanchiwai-ray (approved):

Pjack (approved):

PR #39, opened 2024-08-05
Add support for arm64 builds for charm-local-users

samuelallan72:
This ports the release.yaml changes from charm-local-users to here. :)

### DIFF
--- a/terraform-plans/templates/github/charm_check.yaml.tftpl
+++ b/terraform-plans/templates/github/charm_check.yaml.tftpl
@@ -64,6 +64,7 @@ jobs:
         with:
           provider: "lxd"
           juju-channel: $${{ matrix.juju-channel }}
+          charmcraft-channel: "2.x/stable"
 
       - name: Install latest tox version
         run: |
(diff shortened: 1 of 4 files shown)


samuelallan72:
From my refactor https://github.com/canonical/solutions-engineering-automation/pull/38 :

> type any will be required if we want template variables of other types - for example a list of architectures to run on (useful for multiarch releases coming soon)

Unfortunately it turned out this didn't actually work, so we need to use string values for everything with the current implementation anyway.

jneo8 (approved):

terraform-plans/templates/github/charm_check.yaml.tftpl
```diff
@@ -64,6 +64,7 @@ jobs:
         with:
           provider: "lxd"
           juju-channel: $${{ matrix.juju-channel }}
+          charmcraft-channel: "2.x/stable"
```
chanchiwai-ray, line 67: is it really called `2.x/stable`?

chanchiwai-ray (approved):
A question, otherwise LGTM

terraform-plans/templates/github/charm_check.yaml.tftpl
samuelallan72: Yes

```
$ snap info charmcraft  | grep 2.x/stable
  2.x/stable:       2.7.1                 2024-07-23 (4162) 52MB classic
```

PR #41, opened 2024-08-09
Add config for charmed-openstack-exporter-snap

gabrielcocenza:
This moves the following files for https://github.com/canonical/charmed-openstack-exporter-snap under control of this automation repo:

- CODEOWNERS
- workflows/promote.yaml

It will also apply our standard config to the
charmed-openstack-exporter-snap repository itself.

### DIFF
--- a/terraform-plans/configs/charmed-openstack-exporter-snap_main.tfvars
+++ b/terraform-plans/configs/charmed-openstack-exporter-snap_main.tfvars
@@ -1,0 +1,23 @@
+repository             = "charmed-openstack-exporter-snap"
+repository_description = "Snap package for the OpenStack exporter"
+branch                 = "main"
+workflow_files = {
+  codeowners = {
+    source      = "./templates/github/CODEOWNERS.tftpl"
+    destination = ".github/CODEOWNERS"
+    variables   = {}
+  }
+  promote = {
+    source      = "./templates/github/snap_promote.yaml.tftpl"
+    destination = ".github/workflows/promote.yaml"
+    variables   = {}
+  }
+  release = {
+    source      = "./templates/github/snap_release.yaml.tftpl"
+    destination = ".github/workflows/release.yaml"
+    variables = {
+      branch = "main"
+      include_check_job = false
+    }
+  }
+}
--- a/terraform-plans/configs/charmed-openstack-exporter-snap_main.tfvars
+++ b/terraform-plans/configs/charmed-openstack-exporter-snap_main.tfvars
@@ -1,23 +1,15 @@
 repository             = "charmed-openstack-exporter-snap"
 repository_description = "Snap package for the OpenStack exporter"
 branch                 = "main"
-workflow_files = {
+templates = {
   codeowners = {
     source      = "./templates/github/CODEOWNERS.tftpl"
     destination = ".github/CODEOWNERS"
-    variables   = {}
+    vars   = {}
   }
   promote = {
     source      = "./templates/github/snap_promote.yaml.tftpl"
     destination = ".github/workflows/promote.yaml"
-    variables   = {}
-  }
-  release = {
-    source      = "./templates/github/snap_release.yaml.tftpl"
-    destination = ".github/workflows/release.yaml"
-    variables = {
-      branch = "main"
-      include_check_job = false
-    }
+    vars   = {}
   }
 }
--- a/terraform-plans/templates/github/charm_release.yaml.tftpl
+++ b/terraform-plans/templates/github/charm_release.yaml.tftpl
@@ -7,7 +7,7 @@ name: Release to Edge
 
 on:
   push:
-    branches: [ ${branch} ]
+    branches: [ main ]
   release:
     types: [ published ]
   workflow_dispatch:
--- a/terraform-plans/templates/github/charm_release.yaml.tftpl
+++ b/terraform-plans/templates/github/charm_release.yaml.tftpl
@@ -10,7 +10,6 @@ on:
     branches: [ main ]
   release:
     types: [ published ]
-  workflow_dispatch:
 
 jobs:
   check:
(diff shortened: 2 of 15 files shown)


terraform-plans/configs/charmed-openstack-exporter-snap_main.tfvars
```diff
@@ -0,0 +1,23 @@
+repository             = "charmed-openstack-exporter-snap"
+repository_description = "Snap package for the OpenStack exporter"
+branch                 = "main"
+workflow_files = {
```
samuelallan72, line 4: This was renamed to `templates`.  See other tfvars files for examples. :)

terraform-plans/configs/charmed-openstack-exporter-snap_main.tfvars
```diff
@@ -0,0 +1,23 @@
...
+    source      = "./templates/github/snap_promote.yaml.tftpl"
+    destination = ".github/workflows/promote.yaml"
+    variables   = {}
+  }
+  release = {
+    source      = "./templates/github/snap_release.yaml.tftpl"
+    destination = ".github/workflows/release.yaml"
+    variables = {
+      branch = "main"
+      include_check_job = false
```
samuelallan72, line 20: I don't think this will work; with the structure we have, terraform will enforce all values here to be the same type (cannot mix boolean and string).

terraform-plans/templates/github/snap_release.yaml.tftpl
```diff
@@ -10,15 +10,19 @@ on:
...
   check:
     uses: ./.github/workflows/check.yaml
     secrets: inherit
-
+{{ endif }}
   release:
     runs-on: ubuntu-latest
+{{ if include_check_job }}
     needs: check
+{{ endif }}
```
samuelallan72, line 25: I would prefer to avoid conditionals in templates.  Can we avoid this by adding a check.yaml workflow to charmed-openstack-exporter-snap?  That would be a win-win then.  It only needs to verify that building the snap succeeds.

samuelallan72 (changes requested):

terraform-plans/templates/github/snap_release.yaml.tftpl
```diff
@@ -10,15 +10,19 @@ on:
     branches: [ ${branch} ]
   release:
     types: [ published ]
+  workflow_dispatch:
```
chanchiwai-ray, line 13: If you include `workflow_dispatch`, then should it make more sense to force `check`? Because one can do a release without `check`, that's not ideal

gabrielcocenza: In the case of [charmed-openstack-exporter-snap](https://github.com/canonical/charmed-openstack-exporter-snap) there is no `check` to do because it's just a wrapper to package the code from upstream.

gabrielcocenza: In any case, I've removed the if conditions and now it always will check first.

terraform-plans/templates/github/snap_release.yaml.tftpl
gabrielcocenza: Ok. I think the workflow for the charmed-openstack-exporter is a corner case and not sure if worth adding a new template for it. So I'll try to keep simple and kust have the `promote` and `CODEOWNERS`. If we see a need in the future, we can open a new PR

samuelallan72: I don't think it's a corner case.  I think it's just as useful to have a check workflow as our other projects.  Please see my [other comment](https://github.com/canonical/solutions-engineering-automation/pull/41#discussion_r1714375910) for useful checks to include in the check workflow.

terraform-plans/templates/github/snap_release.yaml.tftpl
chanchiwai-ray: Thanks! I didn't notice that there's only a `snapcraft.yaml` in it. In any case, I think corner case will happen in the future, it's better to have some general pattern to fix it

chanchiwai-ray:
> Error: could not find a branch protection rule with the pattern 'main'

I vaguely remember github terraform cannot set this if you did not set it first cc @jneo8

chanchiwai-ray (approved):

terraform-plans/templates/github/snap_release.yaml.tftpl
samuelallan72: Ok, we can look into adding the release.yaml to automation later :+1:

samuelallan72 (approved):

terraform-plans/templates/github/charm_release.yaml.tftpl
```diff
@@ -10,6 +10,7 @@ on:
     branches: [ main ]
   release:
     types: [ published ]
+  workflow_dispatch:
```
samuelallan72, line 13: @gabrielcocenza actually can we add these in a separate PR?  They aren't related to adding charmed-openstack-exporter-snap.

terraform-plans/templates/github/snap_release.yaml.tftpl
```diff
@@ -10,12 +10,12 @@ on:
     branches: [ main ]
   release:
     types: [ published ]
+  workflow_dispatch:
 
 jobs:
   check:
     uses: ./.github/workflows/check.yaml
     secrets: inherit
-
```
samuelallan72, line 18: Same for changes in this file; they should be dropped in this PR.

samuelallan72 (changes requested):

samuelallan72 (approved):

terraform-plans/templates/github/snap_release.yaml.tftpl
samuelallan72: @gabrielcocenza thanks for addressing the other comments, it looks fine to me now.  I still argue that we should have a test workflow, so I opened https://github.com/canonical/charmed-openstack-exporter-snap/issues/5 to track it.

jneo8 (approved):

PR #56, opened 2024-08-21
Fix the release file for charms

gabrielcocenza:
- currently the file is broken because it's missing brackets

E.g: [0](https://github.com/canonical/hardware-observer-operator/actions/runs/10497388141/workflow)

### DIFF
--- a/terraform-plans/templates/github/charm_release.yaml.tftpl
+++ b/terraform-plans/templates/github/charm_release.yaml.tftpl
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: ${runs_on}
+        runs-on: [${runs_on}]
     steps:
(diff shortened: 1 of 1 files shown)


aieri (approved):

zxhdaze (approved):

samuelallan72 (changes requested):
No, this needs to be fixed by editing the runs_on var for the specific tfvars for affected projects.  Otherwise it's confusing if you need to pass a list because the string var would need to look like `'"ubuntu-latest", ["self-hosted", "arm64"]'` for example (missing brackets).

This change would also only fix the release.yaml for hardware-observer-operator; it will break the release.yaml for all the other projects in automation.

samuelallan72:
Superceded by https://github.com/canonical/solutions-engineering-automation/pull/57 - please see the fix there instead.

PR #58, opened 2024-08-26
Add remaining repositories to automation

samuelallan72:
- minimal github repo settings
- codeowners file


terraform-plans/configs/charmed-openstack-exporter-snap_main.tfvars
chanchiwai-ray, line 1: Can we not change this file if we are only adding remaining repository to automation repo?

chanchiwai-ray (changes requested):
Thanks, I have one question

terraform-plans/configs/charmed-openstack-exporter-snap_main.tfvars
samuelallan72: Good catch, I don't know how that got changed :thinking:

samuelallan72: Updated :)

chanchiwai-ray (approved):
LGTM, however, I do notice `homepage_url` in the github repo is overrided to `null` (e.g. openstack-exporter-operator). This is not a big problem since we don't have many project specify a `homepage_url`, but flagged a potential issue to me: it's hard to review what's being changes to the repo setting without looking at the workflow result.

.github/workflows/terraform-apply.yaml
```diff
@@ -24,17 +24,50 @@ jobs:
         matrix:
           repository:
             # The naming convention of the files is $REPO_$BRANCH.tfvars
+            - bootstack-actions_main
```
Pjack, line 27: Bootstack-actions will be deprecated. But it's not a big issue to apply to this project too.  I am fine with that.

Pjack (approved):

.github/workflows/terraform-apply.yaml
samuelallan72: Thanks :)

PR #69, opened 2024-09-03
Pin all workflows to explicitly run on Jammy

aieri:
With the release of 24.04.1, it's only a matter of time before GitHub
switches the ubuntu-latest image to use Noble. This change pins all
workflows to 22.04 (which is the current behavior) so that we can switch
from Jammy to Noble at our own pace.


samuelallan72 (approved):
LGTM, thanks

chanchiwai-ray (approved):

Issue #63, opened 2024-08-28
Automated PRs should be updated if possible


samuelallan72:
Currently, the workflow to apply changes simply opens new PRs with the changes, even if there are already open pending PRs to target repositories.  It would be better if it detecting that a PR was already open, and updated that PR instead of creating a new one.

samuelallan72:
Note: a fix might be simple - use a consistent hardcoded branch name for PRs.

chanchiwai-ray:
> Note: a fix might be simple - use a consistent hardcoded branch name for PRs.

we can test with a test repo using PAT + terraform + some PRs

PR #74, opened 2024-09-10
Ensure automated PRs are created, not duplicated

samuelallan72:
Previously there was the issue where a new PR to target repositories was created each time the terraform apply workflow runs, if there were outstanding changes to files.
This caused a lot of noise.

By using a consistent branch name and some extra logic,
we can ensure existing PRs are updated,
rather than creating new ones.

Fixes: #63

### DIFF
--- a/terraform-plans/modules/GitHub/templates/main.tf
+++ b/terraform-plans/modules/GitHub/templates/main.tf
@@ -61,7 +61,7 @@ resource "github_branch" "managed_files_branch" {
   count = length(local.changed_files) > 0 ? 1 : 0
 
   repository    = var.repository
-  branch        = "${var.pr_branch}-${random_string.update_uid.id}"
+  branch        = var.pr_branch
   source_branch = var.branch
 }
 
@@ -79,9 +79,27 @@ resource "github_repository_file" "managed_files" {
   depends_on = [github_branch.managed_files_branch]
 }
 
-# Create a pull request only if there are changed files
+data "github_repository_pull_requests" "open" {
+  base_repository = var.repository
+  base_ref        = var.branch
+  state           = "open"
+}
+
+locals {
+    pr_exists = length([for pr in data.github_repository_pull_requests.open.results : pr if pr.head_ref == var.pr_branch]) > 0
+}
+
+locals {
+    existing_pr = local.pr_exists ? [for pr in data.github_repository_pull_requests.open.results : pr if pr.head_ref == var.pr_branch][0] : null
+}
+
+locals {
+    should_create_pr = length(local.changed_files) > 0 && !local.pr_exists
+}
+
+# Create a pull request only if there are changed files and there isn't already a PR open for the branch
 resource "github_repository_pull_request" "managed_files_update_pr" {
-  count = length(local.changed_files) > 0 ? 1 : 0
+  count = local.should_create_pr ? 1 : 0
 
   base_repository = var.repository
   base_ref        = var.branch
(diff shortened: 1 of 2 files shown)


samuelallan72:
The method of simply hardcoding the branch for the PR does work, but the terraform resource to create the PR fails:

> Error: POST https://api.github.com/repos/canonical/charm-nginx/pulls: 422 Validation Failed [{Resource:PullRequest Field: Code:custom Message:A pull request already exists for canonical:automation/update-managed-files.}]

samuelallan72:
It succeeds if the target PR is closed.

Pjack (dismissed):

chanchiwai-ray:
will it update the PR or just don't create new PR?

samuelallan72:
> will it update the PR or just don't create new PR?

@chanchiwai-ray if there is no existing PR, it will create it.  If there is an existing PR, it will update it.

chanchiwai-ray (approved):

terraform-plans/modules/GitHub/templates/main.tf
```diff
@@ -61,7 +61,7 @@ resource "github_branch" "managed_files_branch" {
   count = length(local.changed_files) > 0 ? 1 : 0
 
   repository    = var.repository
-  branch        = "${var.pr_branch}-${random_string.update_uid.id}"
+  branch        = var.pr_branch
```
rgildein, line 64: You did not remove random_string (line 15-20).

rgildein (dismissed):
What if we open two PR here and one is approved -> new PR for targeted repos is created and before such PR (in targeted repo) is approved and merge, another PR in this repo is merged.
I think that this will cause silently not doing any new PR and missing those changes. 
Can we instead of random branch name use some id/hash/... from git commit? e.g. `automation/update-managed-file/dd1bc07`

terraform-plans/modules/GitHub/templates/main.tf
samuelallan72: Thanks, updated.

samuelallan72:
> What if we open two PR here and one is approved -> new PR for targeted repos is created and before such PR (in targeted repo) is approved and merge, another PR in this repo is merged. I think that this will cause silently not doing any new PR and missing those changes.

Sorry I can't follow what you're saying here.

If there's an already open PR, that PR will be updated with new commits.  So we won't miss the changes; we'll see them when reviewing the target repo automated PR.

chanchiwai-ray (approved):

Pjack (approved):

Pjack:
> > What if we open two PR here and one is approved -> new PR for targeted repos is created and before such PR (in targeted repo) is approved and merge, another PR in this repo is merged. I think that this will cause silently not doing any new PR and missing those changes.
> 
> Sorry I can't follow what you're saying here.
> 
> If there's an already open PR in the target repo, that PR will be updated with new commits. So we won't miss the changes; we'll see them when reviewing the target repo automated PR.

@rgildein 
I don't fully understand your wording, but I assume your main concern is that some changes might be merged without review. This could happen if the changes are made after the automated PR has been approved but not merged yet.
We've enabled the 'Dismiss stale pull request approvals when new commits are pushed' setting in every repository, so this issue should no longer happen.

https://github.com/canonical/solutions-engineering-automation/blob/main/terraform-plans/modules/GitHub/settings/main.tf#L82

rgildein:
> > What if we open two PR here and one is approved -> new PR for targeted repos is created and before such PR (in targeted repo) is approved and merge, another PR in this repo is merged. I think that this will cause silently not doing any new PR and missing those changes.
> 
> Sorry I can't follow what you're saying here.
> 
> If there's an already open PR in the target repo, that PR will be updated with new commits. So we won't miss the changes; we'll see them when reviewing the target repo automated PR.

I see, so one PR could have multiple changes from multiple automated commits? If yes, is this what we want? Since what I know, we are doing squash and merge, so multiple changes will be done by single PR <==> single commit. Maybe it's fine, since these changes are done by automation. Not Sure.

samuelallan72:
@rgildein 

> I see, so one PR could have multiple changes from multiple automated commits? If yes, is this what we want?

Yep that's correct. :)  It should be fine, because it's automated - the commit messages and description of the PR doesn't tell us much already; the real change history is in this automation repo.

rgildein:
> Yep that's correct. :) It should be fine, because it's automated - the commit messages and description of the PR doesn't tell us much already; the real change history is in this automation repo.

I agree :+1:, originally I thought that if PR already exists, new changes will not be applied. Thanks!

PR #76, opened 2024-09-11
Inline lint and unit workflows

samuelallan72:
This reduces overall complexity,
makes it easier to debug workflows,
and removes dependency on bootstack-actions.

This is being tested at [canonical/charm-userdir-ldap#42](https://github.com/canonical/charm-userdir-ldap/pull/42)

### DIFF
--- a/terraform-plans/templates/github/charm_check.yaml.tftpl
+++ b/terraform-plans/templates/github/charm_check.yaml.tftpl
@@ -20,17 +20,60 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lint-unit:
-    uses: canonical/bootstack-actions/.github/workflows/lint-unit.yaml@v2
+  lint:
+    name: Lint
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.8", "3.10"]
-    with:
-      python-version: $${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


jneo8 (approved):

terraform-plans/templates/github/charm_check.yaml.tftpl
```diff
@@ -20,17 +20,60 @@ concurrency:
...
+
+    - name: Set up Python $${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: $${{ matrix.python-version }}
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install 'tox<5'
```
rgildein, line 68: I think it would be good to add note why we are pining tox <5. WDYT?

samuelallan72: Sure, good idea, I can do that :+1: 

For the record, the reason is to avoid a repeat of what happened to our CI on every project when tox 4 was originally released.  Safer to pin to a major version, then we can upgrade in a controlled manner when a new major version is released. :sweat_smile:

samuelallan72: Added :)

jneo8 (approved):

Pjack (approved):

PR #78, opened 2024-09-11
Add matrix to runs-on in snaps

gabrielcocenza:
This way it's possible to build and release snaps using arm64 for the projects.

Added arm64 build and release for the dcgm snap

### DIFF
--- a/terraform-plans/configs/dcgm-snap_main.tfvars
+++ b/terraform-plans/configs/dcgm-snap_main.tfvars
@@ -11,7 +11,7 @@ templates = {
     source      = "./templates/github/snap_check.yaml.tftpl"
     destination = ".github/workflows/check.yaml"
     vars        = {
-      runs_on = "[[ubuntu-22.04], [Ubuntu_ARM64_4C_16G_01]]",
+      runs_on = "[[ubuntu-22.04], [self-hosted, jammy, ARM64]]",
     }
   }
   promote = {
@@ -23,7 +23,7 @@ templates = {
     source      = "./templates/github/snap_release.yaml.tftpl"
     destination = ".github/workflows/release.yaml"
     vars        = {
-       runs_on = "[[ubuntu-22.04], [Ubuntu_ARM64_4C_16G_01]]",
+       runs_on = "[[ubuntu-22.04], [self-hosted, jammy, ARM64]]",
     }
   }
   yamllint = {
--- a/terraform-plans/configs/dcgm-snap_main.tfvars
+++ b/terraform-plans/configs/dcgm-snap_main.tfvars
@@ -10,7 +10,9 @@ templates = {
   check = {
     source      = "./templates/github/snap_check.yaml.tftpl"
     destination = ".github/workflows/check.yaml"
-    vars        = {}
+    vars        = {
+      runs_on = "[[ubuntu-22.04], [Ubuntu_ARM64_4C_16G_01]]",
+    }
   }
   promote = {
     source      = "./templates/github/snap_promote.yaml.tftpl"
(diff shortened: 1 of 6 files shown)


aieri (approved):

terraform-plans/configs/dcgm-snap_main.tfvars
```diff
@@ -10,7 +10,9 @@ templates = {
   check = {
     source      = "./templates/github/snap_check.yaml.tftpl"
     destination = ".github/workflows/check.yaml"
-    vars        = {}
+    vars        = {
+      runs_on = "[[ubuntu-22.04], [Ubuntu_ARM64_4C_16G_01]]",
```
samuelallan72, line 14: It's hard to see if this works without merging; could you test this by opening a draft PR to dcgm-snap with the changes to verify that the workflow passes? :)

samuelallan72 (changes requested):
This looks great thanks, but it would be good to see this tested on dcgm-snap before it's merged here. :)

Pjack:
@gabrielcocenza Why we need to support ARM64 for COU project ?

gabrielcocenza:
> Why we need to support ARM64 for COU project ?

@Pjack  We don't. I'm just adding arm64 for `DCGM`. The other projects I'm adding a variable to be populated with what to run. In case of COU, tailscale and etc it's going to run just on amd64

Pjack:
> > Why we need to support ARM64 for COU project ?
> 
> @Pjack We don't. I'm just adding arm64 for `DCGM`. The other projects I'm adding a variable to be populated with what to run. In case of COU, tailscale and etc it's going to run just on amd64

Understand. 

This is another example of why we need to implement hierarchical variables. Ideally, modifying `charmed-openstack-upgrader_main.tfvars` should be unnecessary when we only want to adjust the behavior in `dcgm-snap_main.tfvars`

terraform-plans/configs/dcgm-snap_main.tfvars
```diff
@@ -20,7 +22,9 @@ templates = {
   release = {
     source      = "./templates/github/snap_release.yaml.tftpl"
     destination = ".github/workflows/release.yaml"
-    vars        = {}
+    vars        = {
+       runs_on = "[[ubuntu-22.04], [Ubuntu_ARM64_4C_16G_01]]",
```
chanchiwai-ray, line 26: `Ubuntu_ARM64_4C_16G_02` ?

terraform-plans/configs/dcgm-snap_main.tfvars
gabrielcocenza: That was a good suggestion. I had no idea how many things I had to fix until is finally working :grimacing: 

Take a look at https://github.com/canonical/dcgm-snap/pull/15

terraform-plans/configs/dcgm-snap_main.tfvars
gabrielcocenza: I've changed to use the self-hosted while the IS give us access to the arm runner. But yes, that should be changed to `Ubuntu_ARM64_4C_16G_02` later.

@chanchiwai-ray do you need to start running arm64 on smartctl also? If you need I can include at this PR

terraform-plans/configs/dcgm-snap_main.tfvars
samuelallan72: Wow!  Nice work :grin: I vote we merge https://github.com/canonical/dcgm-snap/pull/15 now that it's passing CI; then when we merge this, we can verify that it's produced the same result easily. :)

samuelallan72 (approved):

jneo8 (approved):

terraform-plans/configs/dcgm-snap_main.tfvars
gabrielcocenza: Now that is merged, I'm solving this thread. Thanks Sam!

terraform-plans/configs/dcgm-snap_main.tfvars
gabrielcocenza: Solving this thread once is using now the self-hosted. We can change once IS gives permission to use the GitHub runners and you can open a new PR when you want to add arm64 for smartctl.

PR #86, opened 2024-09-17
Pin ubuntu to jammy for lint and unit workflows

samuelallan72:
We have it pinned elsewhere, so this makes it consistent.


Pjack (approved):

samuelallan72:
Going to avoid merging this for now until we fix the bugs in the PR logic.

gabrielcocenza (approved):

Issue #88, opened 2024-09-19
Fix issue with possibly wrong base branch for automated PRs


samuelallan72:
see https://github.com/canonical/snap-tempest/pull/246 for example

samuelallan72:
I investigated this and I can't find anything to suggest it's using an incorrect base branch.  It was working fine before https://github.com/canonical/solutions-engineering-automation/pull/74/files , but the only real thing that changed there was reusing the pr branch.  So perhaps a bug in the github terraform provider that's mangling things when it reuses a branch?  Or strange behaviour around adding commits to an existing branch?

PR #89, opened 2024-09-19
Stop PRs to same repo interfering with each other

samuelallan72:
Previously the branch name was static,
which caused issues when terraform opens PRs to multiple target branches to the same repository.
This is because the terraform run for each target branch was trying to work on the same branch,
and opening PRs from the same branch.
This was observed in repos when we maintain multiple branches (eg. snap-tempest with main, stable/caracal, stable/bobcat, etc.).

This fixes it by including the target branch name in the PR branch, ensuring that each PR in a target repository will use its own branch, and there won't be any more cross interference.

Fixes: #88

### DIFF
--- a/terraform-plans/modules/GitHub/templates/main.tf
+++ b/terraform-plans/modules/GitHub/templates/main.tf
@@ -49,12 +49,16 @@ locals {
   }
 }
 
+locals {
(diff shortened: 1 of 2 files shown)


Pjack (approved):

jneo8 (approved):
Nice found! Thanks.

PR #90, opened 2024-09-19
Pin tox and set python version for running the CI on snaps

gabrielcocenza:
- snaps were missing the pinning on tox that can cause workflow breaks on major versions as it happens with charms
- setting python3.10 as the standard for running lint


Deezzir (changes requested):
Let's add the Python lint check as we discussed

Deezzir (approved):
Thanks

samuelallan72 (approved):

PR #97, opened 2024-10-01
Add arm64 support to smartctl-exporter workflow

samuelallan72:


Pjack (approved):

jneo8 (approved):

gabrielcocenza (commented):
We also need to change the snap_release.yaml.tfpl in order to fix what we saw in dcgm. Do you mind doing it in this PR? Otherwise it will overwrite with wrong template

gabrielcocenza:
I've created the https://github.com/canonical/solutions-engineering-automation/pull/98 to fix it in a separate PR. Let's merge it before this one

samuelallan72:
The merge order shouldn't matter, because the auto PR will just get updated. :)

PR #100, opened 2024-10-08
Format terraform related code according to terraform's convention

chanchiwai-ray:
Ran `terraform fmt -recursive`


samuelallan72 (dismissed):
Thanks! :)

Pjack (approved):

samuelallan72 (approved):

PR #108, opened 2024-10-21
Add SSDLC - Vulnerability Response

gabrielcocenza:
Add the SECURITY.md file


gabrielcocenza:
This is a POC PR to add the SECURITY.md file at the dcgm snap repo. IMO we can merge this and if it behaves as expected we can add it to all the other projects

terraform-plans/templates/github/SECURITY.md.tftpl
```diff
@@ -0,0 +1,24 @@
...
+
+# Security policy
+
+If the vulnerability affects a dependency, a new version of the component including the updated
+dependency will be released in the respective store, meaning that no new feature will be included:
+the update will be built on top of the previously last released stable version.
+
+If the vulnerability affects our charm/snap code itself, a new version will be built including the
+security fix on top of the current main branch, meaning that the security update will potentially
+include new previously unreleased features.
```
samuelallan72, line 15: Is this a new policy, or existing? If the latter, could you link to our docs on this? If the former, could we discuss it further? :)

samuelallan72 (dismissed):

terraform-plans/templates/github/SECURITY.md.tftpl
gabrielcocenza: [This](https://discourse.canonical.com/t/ssdlc-v1-1-is-here/4347) is the discourse post asking for it and this is an [internal](https://docs.google.com/document/d/1Gjlfe5M-X2ZgGFoKQ7uUhBpCrKhUCVuljK3xK2_RXiY/edit?tab=t.0#heading=h.esvwnqpeu35u) document (already approved) and [this](https://docs.google.com/document/d/1RolsiWCWhM-sVmHFjLt-Mq5yKcgQJ8GLXssGJREjSL0/edit?tab=t.0) is the spec

samuelallan72: Ah excellent! Could you add these links to the commit message and PR description so we can reference the source later?

terraform-plans/templates/github/SECURITY.md.tftpl
```diff
@@ -0,0 +1,24 @@
+<!-- This file is centrally managed as a template file in https://github.com/canonical/solutions-engineering-automation -->
+<!-- To update the file: -->
+<!-- - Edit it in the canonical/solutions-engineering-automation repository. -->
+<!-- - Open a PR with the changes. -->
+<!-- - When the PR merges, the soleng-terraform bot will open a PR to the target repositories with the changes. -->
+
+# Security policy
+
+If the vulnerability affects a dependency, a new version of the component including the updated
```
Deezzir, line 9: Maybe `project` instead of the component, it sounds like you referring to the dependency

terraform-plans/templates/github/SECURITY.md.tftpl
gabrielcocenza: Done

terraform-plans/templates/github/SECURITY.md.tftpl
samuelallan72: If we reword it here, does it also need to be updated in our spec? Perhaps changes need to go through the spec first?

terraform-plans/templates/github/SECURITY.md.tftpl
samuelallan72: Thanks :)

terraform-plans/templates/github/SECURITY.md.tftpl
gabrielcocenza: The spec is already approved by the security team, so I don't see the need of it

samuelallan72: Agreed, please revert the re-word; we can discuss changes to the wording on the spec later. :)

Deezzir (dismissed):
LGTM

terraform-plans/templates/github/SECURITY.md.tftpl
```diff
@@ -0,0 +1,24 @@
...
+
+# Security policy
+
+If the vulnerability affects a dependency of the project, a new version of the component including
+the updated dependency will be released in the respective store, meaning that no new feature will
+be included: the update will be built on top of the previously last released stable version.
+
+If the vulnerability affects our charm/snap code itself, a new version will be built including the
+security fix on top of the current main branch, meaning that the security update will potentially
+include new previously unreleased features.
```
aieri, line 15: while this is our default overall statement, I find it misleading when applied to a specific security.md file in a project, especially one like dcgm. For example, it may seem to imply that we commit to rebuilding dcgm with updated dependencies when those dependencies carry bugfixes, while we are only mantaining the snap packaging and would therefore only bump the dcgm (or dcgm exporter) version as a whole. I suggest dropping these paragraphs from the default template. If we want to include more specific guidance we should modify this text appropriately for the individual projects.

aieri (changes requested):

terraform-plans/templates/github/SECURITY.md.tftpl
aieri: yes exactly, that initial generic description isn't precise enough to explain what we would do when we act as snap packagers. Which ftr I think should be:
* if a security fix is available in core we rebuild without bumping dcgm
* if a security fix is available in a new dcgm release we bump verisons
* if a security fix is available in a dependency of dcgm but no new dcgm release exists we wait

chanchiwai-ray:
this is stale for some time, any update?

aieri (approved):
Looks good, but we need to add this guidance to all projects.

terraform-plans/templates/github/SECURITY.md.tftpl
samuelallan72: @aieri if this template is generic and to be applied to all projects, where will the specific information that you mentioned live?

Deezzir (approved):

PR #109, opened 2024-10-24
Support charmcraft 3 in openstack-exporter-operator

samuelallan72:
Tested in https://github.com/canonical/openstack-exporter-operator/pull/116


rgildein (approved):
LGTM, there is not much we can test now, but it's necessary to be merged to be able to continue on #113.

chanchiwai-ray (approved):

Issue #111, opened 2024-10-31
Change strategy to create PRs in downstream repos


rgildein:
Part of our initiative of improving central file management is to improve our process in this automatization. The idea is that instead of create PRs in downstream repos after PR in this repo is created, we should create PRs straight forwards.

So instead, the workflow will change the following way:
- Create Draft PR -> CI trigger and user can see the output, without PRs in downstream being created
- convert Draft to regular PR -> PRs in downstream will be created
- reviewer check also PRs in downstream and approve only if workflows do not crash due to some type etc.
- [optional] After PR in this repo is closed, another CI could enable auto-merge in downstream repos and provide approval from the bot. If CI passes, the downstream repos will be merged automatically.

Prerequisite is that all downstream PRs are linked with original repo (the original PR is mention in the downstream, so GitHub would do rest).

related: #52

rgildein:
It was fixed by #112, however for now no auto-merge.

Pjack:
Reopening this issue as the changes in #119 were reverted. We will allocate dedicated time to address it later in this cycle.

samuelallan72:
I like this general flow, but I think we should only be merging PRs generated from main. I think we can do this using the same flow with some care and being sure that it's very clear on the automated PRs which ones are ready for merge or not. Remember with this flow we'll potentially have multiple PRs open to each target repository if we have multiple PRs to automation, and they may or may not conflict.

samuelallan72:
Note that I do strongly see the downsides of only having the downstream PRs created after merging to main here (I've been bitten by merging something with a simple issue, then immediately needing a follow up PR that could have been avoided...). I just think we need to take care to develop a polished alternative to avoid introducing new problems. :sweat_smile:

PR #112, opened 2024-10-31
Update apply logic

rgildein:
With these changes, all PRs in downstream repos will be created, and they will use the same branch name and provide a link to the original repo.

related: #111

PR #113, opened 2024-10-31
Add more projects to the automation

rgildein:
I need to add juju_channels to the configuration and use it in
charm_check template. Add TEST_CHARM env variable to snap_check
template.

Added the check, promote and release for:
 - charm-duplicty
 - charm-promethues-blackbox-exporter
 - charm-prometheus-juju-exporter
 - charm-storage-connector
 - prometheus-backupall-juju-exporter
 - prometheus-juju-exporter
    
related: #111

### DIFF
--- a/terraform-plans/configs/charm-prometheus-juju-exporter_main.tfvars
+++ b/terraform-plans/configs/charm-prometheus-juju-exporter_main.tfvars
@@ -24,7 +24,9 @@ templates = {
   promote = {
     source      = "./templates/github/charm_promote.yaml.tftpl"
     destination = ".github/workflows/promote.yaml"
-    vars        = {}
+    vars        = {
+      charmcraft_channel = "2.x/stable",
+    }
   }
   release = {
     source      = "./templates/github/charm_release.yaml.tftpl"
--- a/terraform-plans/configs/charm-prometheus-juju-exporter_main.tfvars
+++ b/terraform-plans/configs/charm-prometheus-juju-exporter_main.tfvars
@@ -33,6 +33,7 @@ templates = {
     destination = ".github/workflows/release.yaml"
     vars = {
       runs_on = "[[ubuntu-22.04]]",
+      charmcraft_channel = "2.x/stable",
     }
   }
   jira_sync_config = {
--- a/terraform-plans/configs/charm-prometheus-juju-exporter_main.tfvars
+++ b/terraform-plans/configs/charm-prometheus-juju-exporter_main.tfvars
@@ -24,7 +24,7 @@ templates = {
   promote = {
     source      = "./templates/github/charm_promote.yaml.tftpl"
     destination = ".github/workflows/promote.yaml"
-    vars        = {
+    vars = {
       charmcraft_channel = "2.x/stable",
     }
   }
@@ -32,7 +32,7 @@ templates = {
     source      = "./templates/github/charm_release.yaml.tftpl"
     destination = ".github/workflows/release.yaml"
     vars = {
-      runs_on = "[[ubuntu-22.04]]",
+      runs_on            = "[[ubuntu-22.04]]",
       charmcraft_channel = "2.x/stable",
     }
   }
--- a/terraform-plans/configs/charm-prometheus-juju-exporter_main.tfvars
+++ b/terraform-plans/configs/charm-prometheus-juju-exporter_main.tfvars
@@ -12,6 +12,27 @@ templates = {
     destination = ".github/CODEOWNERS"
     vars        = {}
   }
+  check = {
+    source      = "./templates/github/charm_check.yaml.tftpl"
+    destination = ".github/workflows/check.yaml"
+    vars = {
+      runs_on       = "[[ubuntu-22.04]]",
+      test_commands = "['TEST_JUJU3=1 make functional']",
+      juju_channels = "[\"3.4/stable\", \"3.5/stable\"]",
+    }
+  }
+  promote = {
+    source      = "./templates/github/charm_promote.yaml.tftpl"
+    destination = ".github/workflows/promote.yaml"
+    vars        = {}
+  }
+  release = {
+    source      = "./templates/github/charm_release.yaml.tftpl"
+    destination = ".github/workflows/release.yaml"
+    vars = {
+      runs_on = "[[ubuntu-22.04]]",
+    }
+  }
   jira_sync_config = {
     source      = "./templates/github/jira_sync_config.yaml.tftpl"
     destination = ".github/.jira_sync_config.yaml"
--- a/.github/workflows/terraform-apply.yaml
+++ b/.github/workflows/terraform-apply.yaml
@@ -1,4 +1,4 @@
-name: Apply new changes to SolEng repositories
+name: Run terraform
 
 on:
   workflow_dispatch:
--- a/terraform-plans/configs/charm-advanced-routing_main.tfvars
+++ b/terraform-plans/configs/charm-advanced-routing_main.tfvars
@@ -18,6 +18,7 @@ templates = {
     vars = {
       runs_on       = "[[ubuntu-22.04]]",
       test_commands = "['tox -e func']",
+      juju_channels = "[\"3.4/stable\"]",
     }
   }
   promote = {
--- a/terraform-plans/configs/charm-apt-mirror_main.tfvars
+++ b/terraform-plans/configs/charm-apt-mirror_main.tfvars
@@ -18,6 +18,7 @@ templates = {
     vars = {
       runs_on       = "[[ubuntu-22.04]]",
       test_commands = "['tox -e func -- -v --series focal', 'tox -e func -- -v --series jammy']",
+      juju_channels = "[\"3.4/stable\"]",
     }
   }
   promote = {
--- a/terraform-plans/configs/charm-duplicity_main.tfvars
+++ b/terraform-plans/configs/charm-duplicity_main.tfvars
@@ -12,6 +12,27 @@ templates = {
     destination = ".github/CODEOWNERS"
     vars        = {}
   }
+  check = {
+    source      = "./templates/github/charm_check.yaml.tftpl"
+    destination = ".github/workflows/check.yaml"
+    vars = {
+      runs_on       = "[[ubuntu-22.04]]",
+      test_commands = "['TEST_JUJU3=1 make functional']",
+      juju_channels = "[\"3.4/stable\"]",
+    }
+  }
+  promote = {
+    source      = "./templates/github/charm_promote.yaml.tftpl"
+    destination = ".github/workflows/promote.yaml"
+    vars        = {}
+  }
+  release = {
+    source      = "./templates/github/charm_release.yaml.tftpl"
+    destination = ".github/workflows/release.yaml"
+    vars = {
+      runs_on = "[[ubuntu-22.04]]",
+    }
+  }
   jira_sync_config = {
     source      = "./templates/github/jira_sync_config.yaml.tftpl"
     destination = ".github/.jira_sync_config.yaml"
--- a/terraform-plans/configs/charm-duplicity_main.tfvars
+++ b/terraform-plans/configs/charm-duplicity_main.tfvars
@@ -24,7 +24,9 @@ templates = {
   promote = {
     source      = "./templates/github/charm_promote.yaml.tftpl"
     destination = ".github/workflows/promote.yaml"
-    vars        = {}
+    vars        = {
+      charmcraft_channel = "2.x/stable",
+    }
   }
   release = {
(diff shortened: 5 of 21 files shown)


terraform-plans/configs/charm-prometheus-juju-exporter_main.tfvars
```diff
@@ -12,6 +12,27 @@ templates = {
     destination = ".github/CODEOWNERS"
     vars        = {}
   }
+  check = {
+    source      = "./templates/github/charm_check.yaml.tftpl"
+    destination = ".github/workflows/check.yaml"
+    vars = {
+      runs_on       = "[[ubuntu-22.04]]",
+      test_commands = "['TEST_JUJU3=1 make functional']",
+      juju_channels = "['3.4/stable', '3.5/stable']]",
```
samuelallan72, line 21:
```suggestion
      juju_channels = "['3.4/stable', '3.5/stable']",
```

samuelallan72:
@rgildein I've started addressing the changes required for charm-prometheus-blackbox-exporter: https://github.com/canonical/charm-prometheus-blackbox-exporter/pull/21

samuelallan72:
https://github.com/canonical/prometheus-juju-backup-all-exporter/pull/31

samuelallan72:
We'll need to update the template (or use a different one) for https://github.com/canonical/prometheus-juju-exporter because it also has unit tests (the current snap template does not have unit tests).

samuelallan72:
https://github.com/canonical/charm-duplicity/pull/40

Pjack (commented):
Do we have solution to avoid adding identical default value 
`juju_channels = "[\"3.4/stable\"]` 
in every project's tfvar file?

rgildein:
> Do we have solution to avoid adding identical default value `juju_channels = "[\"3.4/stable\"]` in every project's tfvar file?

No, there was discussion and POC from me and @samuelallan72 to avoid terraform, since without states terraform does not make sense. However, in one of our meeting @aieri seems to agree with me that we can use simply remote server for states or commit them to this repo.

Some key points to discussion:
- Do we want to have default values for things like `juju-channel`?
- Do we want to go with simple Jinaja2 templates and GH workflow, which will replace TF for managing files (not settings)?

rgildein:
IMO:
1. We should try to avoid default values, since it's less clear what is actually used. I do not see any issue changing juju-channel in each tfvar maunally, it's one command for me, and it's clear for reviewer what I'm changing. 
2. We spend a lot of time doing (James mentioned 1 month of work) hacks for TF to make it work as we needed, which I believe is a completely wrong approach. We should either save states, where I do not see any security issue, and use the TF as it is intended to be used or go with simple GH workflow running Jinja2 template rendering + create PR.

jneo8:
> > Do we have solution to avoid adding identical default value `juju_channels = "[\"3.4/stable\"]` in every project's tfvar file?
> 
> No, there was discussion and POC from me and @samuelallan72 to avoid terraform, since without states terraform does not make sense. However, in one of our meeting @aieri seems to agree with me that we can use simply remote server for states or commit them to this repo.
> 
> Some key points to discussion:
> 
> * Do we want to have default values for things like `juju-channel`?
> * Do we want to go with simple Jinaja2 templates and GH workflow, which will replace TF for managing files (not settings)?

Can you remind me why we need terraform status now?
My memory stay in we don't need state because we don't want to use terraform to manage the template anymore and remote state also increase the maintenance effort.

chanchiwai-ray:
> Do we want to go with simple Jinaja2 templates and GH workflow, which will replace TF for managing files (not settings)?

I don't remember the exact reason why jinja2 template syntax is better than TF's `template_file`, all I remembered is that we have an issue with `template_file` that all variables must be the same type but that's not a blocker to us, doesn't it?

jneo8:
> IMO:
> 
> 1. We should try to avoid default values, since it's less clear what is actually used. I do not see any issue changing juju-channel in each tfvar maunally, it's one command for me, and it's clear for reviewer what I'm changing.
> 2. We spend a lot of time doing (James mentioned 1 month of work) hacks for TF to make it work as we needed, which I believe is a completely wrong approach. We should either save states, where I do not see any security issue, and use the TF as it is intended to be used or go with simple GH workflow running Jinja2 template rendering + create PR.

- How can you ensure there are no security issues, given that the Terraform state is managed by another company?
- Broadly speaking, the task is simply to compare the generated template file with the file in the remote repository and create a PR if there are differences. Can you explain why a Terraform state is necessary to achieve this, given that we can obtain the two files to compare without using the state?
- It didn’t actually take a full month—the hacking part took less than a week. The entire process, including task planning, testing, hacking, and apply to all the projects spanned two sprints, which is about one month in total.

rgildein:
>     * How can you ensure there are no security issues, given that the Terraform state is managed by another company?

I asked security team, and they do not say that it's dangerous, but they mention that they have no experiences with it.
We can investigate remote server or committing states to this repo. States do not include any secrets and since this repo is public and play are visible in CI, I do not see any differences.

With remote server, we could have an issue if there are multiple PR in this repo, since we are applying in PR now, not after it's merged.

>     * Broadly speaking, the task is simply to compare the generated template file with the file in the remote repository and create a PR if there are differences. Can you explain why a Terraform state is necessary to achieve this, given that we can obtain the two files to compare without using the state?
>       
>       * Also you don't have to spend any effort to maintain the state, which I don't think it's a wrong approach.

We are configuring all repos always, the plan does not show current state. We did stage hacks (compare files) to create PR or not, which goes against of logic of TF. The TF should be capable to know if RP is already there and not pushing same commits again and again, as you can see [here](https://github.com/canonical/charm-duplicity/pull/39).

I believe that we either save TF states (committing here or remote server or ...) or simply not use TF for it at all.

>     * It didn’t actually take a full month—the hacking part took less than a week. The entire process, including task planning, testing, hacking, and apply to all the projects spanned two sprints, which is about one month in total.

:+1:

chanchiwai-ray (dismissed):
Why are we putting all these inactive projects to automation repos?

jneo8:
> I believe that we either save TF states (committing here or remote server or ...) or simply not use TF for it at all.

I will say TF is not a best option for our case.

I really have concern about the using TF state, even it's remote. We need to handle issues like  out of sync and security.

Please consider non-TF solution as first priority, thanks.

samuelallan72:
@rgildein do you have a PoC or example for how the code and workflow would look if we continue to use terraform templates for managing files but add terraform state? I haven't seen an example yet, so I'm not sure exactly how well that will improve the situation.

I'm strongly for dropping terraform for managing the files and PRs - it's not what terraform is designed for and it's really awkward to use and maintain. If we switch to github workflow for managing the PRs along with a script to template the files, it will be much easier to maintain and more flexible.

The PoC I did at the sprint can be found at https://github.com/samuelallan72/test-automation

@rgildein could you link to your PoC from the sprint too?

samuelallan72:
> Why are we putting all these inactive projects to automation repos?

IMHO because these projects are still maintained projects, and keeping the workflows, pipelines, etc. updated makes it easier to maintain the projects.

chanchiwai-ray:
@rgildein I think you will need to drop charm-prometheus-juju-exporter for now, since it might need a new template to support corner case https://github.com/canonical/charm-prometheus-juju-exporter/pull/68

samuelallan72:
> I think you will need to drop charm-prometheus-juju-exporter for now, since it might need a new template to support corner case

Note: ideally we would use the same template for all related projects, except for cases where there is a strong case for a major difference. Otherwise we will end up with many templates and lose the advantage of enforcing consistency.

I agree that for now, with the limitations of our current structure and use of terraform templates, it makes sense to split templates. But longer term, when we switch to a better templating engine, we can definitely ensure we support these cases in a single template - especially in the two recent corner cases involving more complex matrix definition ( https://github.com/canonical/charm-prometheus-juju-exporter/pull/68 and another from @rgildein ) - this should be quite possible to achieve.

rgildein:
> Why are we putting all these inactive projects to automation repos?

To be able to drop bootstack-actions.

> I'm strongly for dropping terraform for managing the files and PRs - it's not what terraform is designed for and it's really awkward to use and maintain.

Why do you think it's not designed for these steps? I think it's not designed to be used without states, and we are using it exactly like that. The benefit of using states is that terraform can know that PR was created with exact commits and not push anything + git configuration will be applying only new changes and not always changing everything.

> @rgildein could you link to your PoC from the sprint too?

https://github.com/rgildein/solutions-engineering-automation/pull/1

> @rgildein I think you will need to drop charm-prometheus-juju-exporter for now, since it might need a new template to support corner case https://github.com/canonical/charm-prometheus-juju-exporter/pull/68

We can merge it with it, and simply closed PR create by this automation and then @chanchiwai-ray can open new PR here with proper template. I do not want to remove it from the list, because we simply forgot to add.

> Note: ideally we would use the same template for all related projects, except for cases where there is a strong case for a major difference. Otherwise we will end up with many templates and lose the advantage of enforcing consistency.

What is an issue to have multiple templates? I think we should avoid to creating some hacks (as we did for bootstack-action) to avoid having multiple templates. These hacks make any maintaining more complicated. The benefit of these automation is not to have singe template, but have single place where we can do simple changes, like `3.4/stable` -> `3.5/stable` for Juju. 

With Jinja, we can do some logic inside templates, however I believe it will simply hide stuff and make it more complex to maintain.

chanchiwai-ray:
> We can merge it with it, and simply closed PR create by this automation and then @chanchiwai-ray can open new PR here with proper template. I do not want to remove it from the list, because we simply forgot to add.

That works too, let me know how we are going to handle these corner case (multiple templates or something Samuel suggested)

samuelallan72:
> What is an issue to have multiple templates? I think we should avoid to creating some hacks (as we did for bootstack-action) to avoid having multiple templates. These hacks make any maintaining more complicated. The benefit of these automation is not to have singe template, but have single place where we can do simple changes, like 3.4/stable -> 3.5/stable for Juju.

> With Jinja, we can do some logic inside templates, however I believe it will simply hide stuff and make it more complex to maintain.

Agreed, I don't want to enforce a single template at the expense of losing simplicity. However, I believe we should strive to have as small a number of templates as possible, to help keep our workflows as consistent as possible. Consistency is the main benefit here (consistency includes having a single place to do simple changes like changing the juju version).

> let me know how we are going to handle these corner case 

My take on this is that for every change that deviates from the template is to consider how much of an edge case it is. If it's something truly radical and not likely to be used by other projects, then go for it and make a new template. However, if it's not really that much of a corner case, then we should consider refactoring the templates to better accomodate these changes. For exmaple, here we're taking about adding a new combination of matrix variables, which I've seen in at least one other PR recently too - so it's not really a strong corner case - here we already use template variables for matrix, so we can refactor how we use variables in matrix to accomodate how we want to structure it. That's the benefit of being able to refactor.

rgildein:
> Agreed, I don't want to enforce a single template at the expense of losing simplicity. However, I believe we should strive to have as small a number of templates as possible, to help keep our workflows as consistent as possible. Consistency is the main benefit here (consistency includes having a single place to do simple changes like changing the juju version).
>

There will be always exceptions, and with thous I believe using single value is not good. This is single repo where you can change it, so I do not think you need also single value. At least that's me.
  
> My take on this is that for every change that deviates from the template is to consider how much of an edge case it is. If it's something truly radical and not likely to be used by other projects, then go for it and make a new template. However, if it's not really that much of a corner case, then we should consider refactoring the templates to better accomodate these changes. For exmaple, here we're taking about adding a new combination of matrix variables, which I've seen in at least one other PR recently too - so it's not really a strong corner case - here we already use template variables for matrix, so we can refactor how we use variables in matrix to accomodate how we want to structure it. That's the benefit of being able to refactor.

I'm ok to do any refactor, but I simply want to avoid doing hacks in original repo in the workflow or here. We can add exclude as parameter for our matrixes here instead of creating new template (but that will require to add empty value to all repos except one). I agree that if change is not radical, why not.

Pjack (approved):
Most of the child PR are pass , we could merge and investigate further issues if any.

samuelallan72 (approved):
Agreed, let's merge and go from there. :)

samuelallan72 (changes requested):
Actually no wait, we need to drop commit https://github.com/canonical/solutions-engineering-automation/pull/113/commits/9638ed6bf0d49841ed10d4677790d2c7a759d6d6

samuelallan72:
I did some updates here, so it should be ready for review again:

- dropped the cherry picked commit from https://github.com/canonical/solutions-engineering-automation/pull/112 (it's been reverted, and needs more discussion and work before it can be reintroduced)
- rebased on main
- updated the test commands for two charms that have had their structure updated recently.

samuelallan72 (approved):

samuelallan72 (approved):

chanchiwai-ray (approved):

PR #119, opened 2024-11-27
Revert "Update apply logic"

Pjack:
Reverts canonical/solutions-engineering-automation#112

### DIFF
--- a/.github/workflows/terraform-apply.yaml
+++ b/.github/workflows/terraform-apply.yaml
@@ -117,12 +117,10 @@ jobs:
           terraform plan -no-color \
           -var-file=configs/github.tfvars \
           -var-file=configs/${{ matrix.repository }}.tfvars \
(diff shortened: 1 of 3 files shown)


samuelallan72 (approved):

jneo8 (approved):
The main reason I approve of this revert is that creating a PR before actually merging it can lead to issues where files become misaligned in the end.

For example, imagine Person A creates a PR that spawns multiple child PRs. At the same time, Person B creates another PR with its own set of child PRs. In such a scenario, it becomes crucial to ensure that there are no conflicts between the two original PRs and that all the child PRs are merged in the correct order.

While the original issue we aimed to address—testing the PR before merging—does get resolved, this approach introduces a new challenge. When multiple people work on the repository simultaneously, there’s a real risk of misalignment, which can complicate the process.

PR #123, opened 2024-12-02
Templated promotion options for snap workflow

Deezzir:
Adds a `promote_options` template var for the `snap_promote` workflow

Closes: #127

### DIFF
--- a/terraform-plans/configs/dcgm-snap_main.tfvars
+++ b/terraform-plans/configs/dcgm-snap_main.tfvars
@@ -23,7 +23,7 @@ templates = {
     source      = "./templates/github/snap_promote.yaml.tftpl"
     destination = ".github/workflows/promote.yaml"
     vars = {
-      promote_options = "['core24/edge -> core24/candidate', 'core24/candidate -> core24/stable']"
+      promote_options = "['latest/edge -> latest/candidate', 'latest/candidate -> latest/stable', 'core24/edge -> core24/candidate', 'core24/candidate -> core24/stable']"
     }
   }
   tics = {
--- a/terraform-plans/configs/dcgm-snap_main.tfvars
+++ b/terraform-plans/configs/dcgm-snap_main.tfvars
@@ -23,12 +23,7 @@ templates = {
     source      = "./templates/github/snap_promote.yaml.tftpl"
     destination = ".github/workflows/promote.yaml"
     vars = {
-      promote_options = [
-        "core24/edge -> core24/candidate",
-        "core24/candidate -> core24/stable",
-        "latest/edge -> latest/candidate",
-        "latest/candidate -> latest/stable"
-      ]
+      promote_options = "['core24/edge -> core24/candidate', 'core24/candidate -> core24/stable']"
     }
   }
   tics = {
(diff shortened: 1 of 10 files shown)


Deezzir:
@aieri Why are you against having these two options for other snaps? It's just another entry in the manual dispatch. It will be up to the developer to choose an appropriate option. It is not run automatically

samuelallan72:
> It's just another entry in the manual dispatch. It will be up to the developer to choose an appropriate option.

Ideally we would make it easy for the developer to choose a valid option though, by only including options valid for the project.

This could be a good candidate for a template variable actually - I guess most of our projects are going to use the same consistent promotion flow, but some projects may have different tracks.

aieri:
> @aieri Why are you against having these two options for other snaps? It's just another entry in the manual dispatch. It will be up to the developer to choose an appropriate option. It is not run automatically

Because I think it's not a good UX to offer options that are known to fail. If I see a drop down menu I expect all options to be valid, which is not the case here (not for all projects).
A free textbox is different because of course the possible options are infinite.

That said, it's also true that the target demographics are developers working on the project, not end users. If you find two +1s that are ok with this design I am not going to block, so I'll dismiss my review.

samuelallan72 (changes requested):
Chiming in to back up @aieri's concern. I agree that we should provide only valid options to the user here - regardless of the demographic, good UX is still a good goal. And I believe we can, with not much more effort. One possibility may be to still use the same template, by using a template variable for these options.

terraform-plans/configs/dcgm-snap_main.tfvars
```diff
@@ -22,7 +22,9 @@ templates = {
   promote = {
     source      = "./templates/github/snap_promote.yaml.tftpl"
     destination = ".github/workflows/promote.yaml"
-    vars        = {}
+    vars = {
+      promote_options = "['core24/edge -> core24/candidate', 'core24/candidate -> core24/stable']"
```
aieri, line 26: should we also include latest/* ? Do we want to keep latest/ and core24/ in sync?

Deezzir: I don't know. I guess we should, does it mean that there are going to be 4 options for dcgm?

samuelallan72 (dismissed):
Thanks, this looks good :+1:

terraform-plans/configs/dcgm-snap_main.tfvars
```diff
@@ -22,7 +22,9 @@ templates = {
   promote = {
     source      = "./templates/github/snap_promote.yaml.tftpl"
     destination = ".github/workflows/promote.yaml"
-    vars        = {}
+    vars = {
```
gabrielcocenza, line 25: Maybe it's not scope of this PR, but because we have this situation of the `core24` I would like to see how we are going to approach this situation.

On release.yaml we are releasing just for `latest/edge`, which means that just accessing the store we would be able to get what is from latest/edge and pass to core24/edge

In summary, I think we should have a new template variable that will be used in a matrix for the releases and in this case releasing on both `core24` and `latest/edge`.

jneo8 (approved):
Thanks, LGTM.

jneo8 (approved):

samuelallan72 (approved):

Issue #115, opened 2024-11-18
Add centrally managed TICS workflows


Deezzir:
Projects:
- `smarctl-exporter-snap` - https://github.com/canonical/smartctl-exporter-snap/pull/18
- `dcgm-exporter-snap` - https://github.com/canonical/dcgm-snap/pull/46
- `charmed-openstack-exporter-snap` - https://github.com/canonical/charmed-openstack-exporter-snap/pull/13

samuelallan72:
@Deezzir the linked PRs have been merged. Are there any blockers to moving these TICS workflows to automation now? :)

Deezzir:
Feel free to take the item, if you want, you guys have more experience with the central management :)

PR #125, opened 2024-12-02
Add TICS automation

Deezzir:
Closes: #115

### DIFF
--- a/terraform-plans/templates/github/snap_tics.yaml.tftpl
+++ b/terraform-plans/templates/github/snap_tics.yaml.tftpl
@@ -43,6 +43,6 @@ jobs:
           mode: qserver
           project: ${project}
           viewerUrl: https://canonical.tiobe.com/tiobeweb/TICS/api/cfg?name=default
-          branchdir: ${{ github.workspace }}
-          ticsAuthToken: ${{ secrets.TICSAUTHTOKEN }}
+          branchdir: $${{ github.workspace }}
+          ticsAuthToken: $${{ secrets.TICSAUTHTOKEN }}
           installTics: true
--- a/terraform-plans/templates/github/snap_tics.yaml.tftpl
+++ b/terraform-plans/templates/github/snap_tics.yaml.tftpl
@@ -27,7 +27,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: snap_artifact
-          path: ${{ steps.build.outputs.snap }}
+          path: $${{ steps.build.outputs.snap }}
 
   tics-report:
     needs: build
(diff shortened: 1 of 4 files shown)


terraform-plans/templates/github/snap_tics.yaml.tftpl
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Verify snap builds successfully
+        id: build
+        uses: canonical/action-build@v1
+
+      - name: Upload the built snap
+        uses: actions/upload-artifact@v4
+        with:
+          name: snap_artifact
+          path: ${{ steps.build.outputs.snap }}
```
samuelallan72, line 30:
```suggestion
          path: $${{ steps.build.outputs.snap }}
```

You'll need to escape the dollar sign when there's a `${` that you don't intend to be processed as a template variable by terraform.

samuelallan72 (changes requested):

terraform-plans/templates/github/snap_tics.yaml.tftpl
```diff
@@ -0,0 +1,68 @@
...
+          fetch-depth: 0
+
+      - name: Verify snap builds successfully
+        id: build
+        uses: canonical/action-build@v1
+
+      - name: Upload the built snap
+        uses: actions/upload-artifact@v4
+        with:
+          name: snap_artifact
```
samuelallan72, line 29: Why do you have a build job here? There is already a build check in the check workflow. If we do really need to use a newly built snap here, then please avoid uploading artefacts - simply build the snap and run the tics target in the same job. :)

samuelallan72 (changes requested):

terraform-plans/templates/github/snap_tics.yaml.tftpl
samuelallan72: (unless of course there's a need to retain the built snap artefact for some other purpose)

terraform-plans/templates/github/snap_tics.yaml.tftpl
```diff
@@ -0,0 +1,48 @@
...
+          sudo apt update
+          python -m pip install --upgrade pip
+          # tics action requires flake8 to be installed on the runner
+          # https://github.com/tiobe/tics-github-action/issues/410
+          python -m pip install flake8
+          # pin tox to the current major version to avoid
+          # workflows breaking all at once when a new major version is released.
+          python -m pip install 'tox<5'
+
+      - name: Verify snap builds successfully
```
samuelallan72, line 33:
```suggestion
      # The snap is built already in the check workflow and available as an artifact,
      # but we want this TICS workflow to be independent,
      # so we can run the workflow periodically or on demand.
      - name: Build the snap
```

It would be nice to document what we discussed earlier - the reason why we want to build the snap again here.

samuelallan72 (approved):

Pjack (approved):

PR #129, opened 2024-12-05
Use tox environment for the TICS workflow 

Deezzir:
Depends on the following PRs to be merged first:

- https://github.com/canonical/smartctl-exporter-snap/pull/23
- https://github.com/canonical/dcgm-snap/pull/51
- https://github.com/canonical/charmed-openstack-exporter-snap/pull/17

### DIFF
--- a/terraform-plans/templates/github/snap_tics.yaml.tftpl
+++ b/terraform-plans/templates/github/snap_tics.yaml.tftpl
@@ -23,25 +23,26 @@ jobs:
         run: |
           sudo apt update
           python -m pip install --upgrade pip
-          # tics action requires flake8 and pylint to be installed on the runner
-          # https://github.com/tiobe/tics-github-action/issues/410
-          python -m pip install flake8 pylint
           # pin tox to the current major version to avoid
           # workflows breaking all at once when a new major version is released.
           python -m pip install 'tox<5'
 
-      # The snap is built already in the check workflow and available as an artifact,
-      # but we want this TICS workflow to be independent,
-      # so we can run the workflow periodically or on demand.
+        # The snap is built already in the check workflow and available as an artifact,
+        # but we want this TICS workflow to be independent,
+        # so we can run the workflow periodically or on demand.
       - name: Build the snap
         id: build
         uses: canonical/action-build@v1
 
       - name: Test with tox and produce coverage report
-        run: tox -e tics
+        run: tox --workdir /tmp/tox -e tics
 
       - name: Run TICS analysis
         uses: tiobe/tics-github-action@v3
+        env:
+          # Add the the tox bin directory to the PATH so that the TICS action can find the tests dependencies.
+          # Other paths are added to the PATH to make sure that the TICS action can find the required binaries.
+          PATH: "/tmp/tox/tics/bin:/snap/bin:/home/runner/.local/bin:/home/runner/.cargo/bin:/bin:/usr/bin:/usr/local/bin:"
         with:
           mode: qserver
           project: ${project}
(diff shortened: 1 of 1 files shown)


terraform-plans/templates/github/snap_tics.yaml.tftpl
```diff
@@ -23,25 +23,26 @@ jobs:
...
+        # The snap is built already in the check workflow and available as an artifact,
+        # but we want this TICS workflow to be independent,
+        # so we can run the workflow periodically or on demand.
       - name: Build the snap
         id: build
         uses: canonical/action-build@v1
 
       - name: Test with tox and produce coverage report
-        run: tox -e tics
+        run: tox --workdir /tmp/tox -e tics
```
jneo8, line 38: **question:** 

Can you leave a comment here to explain why we need this change?

samuelallan72 (commented):
I'm uneasy about this. Why would the tics target require all the same dependencies as the functional tests as well? It feels like merging requirements of the tics tox target, and the tics github action will cause confusion, because it's not a clear separation of concerns for the dependencies.

Can we treat the tics action dependencies separate to the tox target dependencies? Maybe even by creating our own virtualenvironment for the tics action to use.

If we do go down the route of sharing the tox environment, please clearly document with code comments the purpose of the extra dependencies, and note that the tox environment is manually used for the tics github action?

Deezzir:
https://warthogs.atlassian.net/browse/TICSISSUES-3 Please see the issue, unfortunately, it's the way to go. The TICS analysis requires test dependencies to be present to analyze the code and code paths. Additionally, i think it also analyzes deps for the known issues

terraform-plans/templates/github/snap_tics.yaml.tftpl
```diff
@@ -23,25 +23,26 @@ jobs:
...
 
       - name: Run TICS analysis
         uses: tiobe/tics-github-action@v3
+        env:
+          # Add the the tox bin directory to the PATH so that the TICS action can find the tests dependencies.
+          # Other paths are added to the PATH to make sure that the TICS action can find the required binaries.
+          PATH: "/tmp/tox/tics/bin:/snap/bin:/home/runner/.local/bin:/home/runner/.cargo/bin:/bin:/usr/bin:/usr/local/bin:"
```
samuelallan72, line 45:
```suggestion
          PATH: "/tmp/tox/tics/bin:$PATH"
```

Does this work in github workflows? That would save you needing to hardcode the rest of the path

samuelallan72:
@Deezzir ah sorry, thanks for the information.

In that case, it sounds like sharing the tox environment is the way to go. :+1: :+1:

terraform-plans/templates/github/snap_tics.yaml.tftpl
```diff
@@ -23,25 +23,26 @@ jobs:
...
 
       - name: Test with tox and produce coverage report
-        run: tox -e tics
+        run: tox --workdir /tmp/tox -e tics
 
       - name: Run TICS analysis
         uses: tiobe/tics-github-action@v3
+        env:
+          # Add the the tox bin directory to the PATH so that the TICS action can find the tests dependencies.
+          # Other paths are added to the PATH to make sure that the TICS action can find the required binaries.
```
samuelallan72, line 44: Could you also expand this comment please to add the "why"? - ie. why does the tics action need the tests dependencies. The explanation on [TICSISSUES-3](https://warthogs.atlassian.net/browse/TICSISSUES-3) that you linked to was good, but it would also be good to document this here too.

terraform-plans/templates/github/snap_tics.yaml.tftpl
Deezzir: I checked, didn't work, it was missing curl

samuelallan72 (approved):
LGTM. Thanks for adding comments for context. :+1:

PR #130, opened 2024-12-06
Add channels var to  snap_release workflow

Deezzir:
This PR is because we encountered an issue with the DCGM snap. Currently, the default track is `core24`, so we don't automatically push any revisions to that channel. Also, we might want to keep the `latest` in sync.

I looked at the source code of the action we use to release the snaps; this change should work. Check out the change for the dcgm snap - it has a comma-separated list of the channels. And according to the help from the `snapcraft upload`, you can do that:

Action source code: https://github.com/canonical/action-publish/blob/214b86e5ca036ead1668c79afb81e550e6c54d40/src/publish.ts#L54

```
...
Summary:
    By passing --release with a comma-separated list of channels the snap would
    be released to the selected channels if the store review passes for this
    <snap-file>.
...
```

**It should work in theory.**

In addition, I am wondering if I should add additional quotes. Like this: 

```
...
vars = {
      runs_on  = "[[ubuntu-22.04]]",
      channels = "'core24/edge,latest/edge'"
}
...
```

Please let me know if you have any suggestions on how to test the changes other than merging them.

### DIFF
--- a/terraform-plans/configs/charmed-openstack-exporter-snap_main.tfvars
+++ b/terraform-plans/configs/charmed-openstack-exporter-snap_main.tfvars
@@ -31,6 +31,7 @@ templates = {
     destination = ".github/workflows/release.yaml"
     vars = {
       runs_on = "[[ubuntu-22.04], [self-hosted, jammy, ARM64]]",
+      channels = "latest/edge"
     }
   }
   yamllint = {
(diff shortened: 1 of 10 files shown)


samuelallan72 (approved):
Thanks, LGTM! We'll also need this for https://github.com/canonical/prometheus-juju-exporter/pull/91/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaL47-R47

PR #135, opened 2024-12-11
Add SECURITY.md to all projects

gabrielcocenza:
Similar to #108, this adds SECURITY.md to the soleng projects as part of the SSDLC


samuelallan72 (approved):
LGTM

PR #138, opened 2025-01-02
fix: Change the tag for sysconfig functional test runs-on

jneo8:
The old tags runner no longer exists, change to use new tag for `runs-on`.


Pjack (approved):
Confirmed that the format of new tag is correct.

samuelallan72:
> The old tags runner no longer exists...

@jneo8 I can still see the original tags in https://github.com/canonical/charm-sysconfig/actions/runners?tab=self-hosted . Could you explain why this change is required?

PR #139, opened 2025-01-22
Add managed `pyproject.toml` file

Deezzir:
Closes: #82 

I looked at multiple `pyproject.toml` files, and the version I compiled seems reasonable and supports charms and snaps. Please let me know if there is anything I should add/remove/modify

PR #140, opened 2025-02-06
Fix pyproject.toml

gabrielcocenza:
Some fields introduced on #139 are duplicated and turns into an invalid pyproject.toml file.

### DIFF
--- a/terraform-plans/templates/github/pyproject.toml.tftpl
+++ b/terraform-plans/templates/github/pyproject.toml.tftpl
@@ -5,19 +5,30 @@
 # - When the PR merges, the soleng-terraform bot will open a PR to the target repositories with the changes.
 
 [tool.flake8]
-ignore = ["C901", "D100", "D101", "D102", "D103", "W503", "W504"]
-exclude = ['.eggs', '.git', '.tox', '.venv', '.build', 'build', 'lib', 'report']
 max-line-length = 99
 max-doc-length = 99
 max-complexity = 10
-exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv", ".venv"]
+exclude = [
+    ".git",
+    "__pycache__",
+    ".tox",
+    ".build",
+    "build",
+    "dist",
+    ".eggs",
+    "*.egg_info",
+    "venv",
+    ".venv",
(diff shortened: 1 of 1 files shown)


terraform-plans/templates/github/pyproject.toml.tftpl
```diff
@@ -5,19 +5,30 @@
...
+    "lib",
+]
 select = ["E", "W", "F", "C", "N", "R", "D", "H"]
 # Ignore W503, E501 because using black creates errors with this
 # Ignore D107 Missing docstring in __init__
 # Ignore D415 Docstring first line punctuation (doesn't make sense for properties)
 # Ignore N818 Exceptions end with "Error" (not all exceptions are errors)
-ignore = ["W503", "E501", "D107", "D415", "N818"]
 # D100, D101, D102, D103: Ignore missing docstrings in tests
+ignore = ["C901", "W503", "E501", "D107", "D415", "N818", "D100", "D101", "D102", "D103"]
```
Vultaire, line 31: It looks like we dropped an ignore for W504.  We probably want to add this; I would suggest a separate MR for it so we can at least mostly fix this for now.

terraform-plans/templates/github/pyproject.toml.tftpl
```diff
@@ -5,19 +5,30 @@
...
+    ".tox",
+    ".build",
+    "build",
+    "dist",
+    ".eggs",
+    "*.egg_info",
+    "venv",
+    ".venv",
+    "report",
+    "lib",
```
Vultaire, line 23: Wait, are we sure we want to exclude lib?  I know I see built charms often pulling from lib; this is OK if none of our dependent charms have their own prepopulated lib directory, but are we sure of this?

I would be more comfortable if "lib" was not excluded.  I don't have a problem with the other changes here, but I think we should be linting "lib" unless we're certain it's only for modules pulled from external sources.

Vultaire (approved):
I'm going to approve this for the sake of fixing the builds - *but*, I think we need to follow this up with another PR to:

* Remove ```.lib``` from the exclude list for flake8.
* Add back in the "W504" flake8 ignore entry.

PR #141, opened 2025-02-06
Attempted corrections for a few issues found in PR #140

Vultaire:
In #140, I noted a few errata, but as the code was breaking other projects by leaving it as-is, I allowed the short-term merge without changes.

There were 2 notable issues I noted in #140 which concerned me:

* We dropped an ignore clause for pyflake warning W504.
* We added some additional paths to exclude, which I generally was OK with, except for adding the "lib" path.  I'm concerned this may unexpectedly skip flake8 checking in some modules where we may want those checks.

Please let me know what you think.

### DIFF
--- a/terraform-plans/templates/github/pyproject.toml.tftpl
+++ b/terraform-plans/templates/github/pyproject.toml.tftpl
@@ -20,7 +20,6 @@ exclude = [
     "venv",
     ".venv",
     "report",
-    "lib",
 ]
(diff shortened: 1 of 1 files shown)


gabrielcocenza (approved):

Deezzir (approved):
Thanks!

Vultaire:
Amended my commit with a valid GPG signature.  Now merging.

PR #143, opened 2025-02-14
Add var for coverage in pyproject.toml

samuelallan72:
Some projects don't have 100% coverage,
so we can make exceptions here.

Also ignore the docs dir in flake8 config (docs is ignored for other tools, so this makes it consistent).

### DIFF
--- a/terraform-plans/configs/hardware-observer-operator_main.tfvars
+++ b/terraform-plans/configs/hardware-observer-operator_main.tfvars
@@ -55,8 +55,8 @@ templates = {
   pyproject = {
     source      = "./templates/github/pyproject.toml.tftpl"
     destination = "pyproject.toml"
-    vars        = {
-        coverage_threshold_percent = "100"
+    vars = {
+      coverage_threshold_percent = "100"
     }
   }
 }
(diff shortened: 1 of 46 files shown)


terraform-plans/configs/hardware-observer-operator_main.tfvars
```diff
@@ -42,7 +42,7 @@ templates = {
     destination = ".github/.jira_sync_config.yaml"
     vars = {
       component = "hardware-observer",
-      epic_key  = "SOLENG-190"
+      epic_key  = "SOLENG-1100"
```
gabrielcocenza, line 45: Did you change by mistake? I think we can leave the epic as hardware observer

gabrielcocenza (commented):
Instead of a fixed percentage, would be possible to kind of force that new lines added into the project need coverage?

Maybe we could achieve this by reading the content of the report?

terraform-plans/configs/hardware-observer-operator_main.tfvars
samuelallan72: Oh good catch thanks, yep that's a mistake.

samuelallan72:
> Instead of a fixed percentage, would be possible to kind of force that new lines added into the project need coverage?

This is possible - I think it's called diff cover. That could be worth looking into sometime, although I'm not sure how much effort would be required to modify our CI and/or tox/python configs to make it work for us.

Pjack (dismissed):

chanchiwai-ray (approved):

jneo8 (approved):

chanchiwai-ray (approved):

PR #154, opened 2025-02-25
Add common tox.ini for all projects

samuelallan72:
Consolidate differences across all our active repositories into a common tox.ini file that we can apply to them all.


samuelallan72:
There are still some notes and todos inline in the template - I'm interested in feedback at this stage. :)

Also I'm not sure if we can or should use the same tox.ini template for snaps too? I haven't dug deep into the differences, but in general there shouldn't be much difference.

terraform-plans/configs/openstack-exporter-operator_main.tfvars
```diff
@@ -53,4 +53,11 @@ templates = {
       repository = "openstack-exporter-operator"
     }
   }
+  tox = {
```
chanchiwai-ray, line 56: can you apply this template (and also pyproject.toml) to the project first, then port back here

chanchiwai-ray (changes requested):
Since openstack-exporter-operator uses `ruff` which comes by default when initializing with `charmcraft init`, we need special handling for this project. But since projects using `ruff` is a minority, for me it's fine to replace with `pylint`  (and other tools)

terraform-plans/configs/openstack-exporter-operator_main.tfvars
samuelallan72: My plan was to merge this once approved, and then resolve any issues in the target repos on the automated PRs (similar to what we did with pyproject.toml). This won't work out of the box for some projects, but we can work on any refactoring required on the automated PRs.

samuelallan72:
> Since openstack-exporter-operator uses ruff which comes by default when initializing with charmcraft init, we need special handling for this project.

My plan was to not do any special handling - to simply switch to using the same lints as the other projects in the process.

terraform-plans/configs/openstack-exporter-operator_main.tfvars
chanchiwai-ray: Got it

terraform-plans/templates/github/charm_tox.ini.tftpl
```diff
@@ -0,0 +1,102 @@
+[tox]
```
chanchiwai-ray, line 1: Please add comment saying this is a centralized managed file

samuelallan72: Thanks, added :)

terraform-plans/configs/openstack-exporter-operator_main.tfvars
samuelallan72: (I'm not sure if this is the best way?) Either way, it would be good to agree on a template first, and then we have a standard base to apply to all the repositories.

samuelallan72:
I looked through the tox.ini files for the snaps and packages too, and they aren't that different, so I added them too.

terraform-plans/templates/github/tox.ini.tftpl
```diff
@@ -0,0 +1,95 @@
...
+[testenv:func]
+%{ if functest_type == "none" ~}
+allowlist_externals =
+    echo
+commands = echo "No func tests, skipping."
+%{ else ~}
+deps =
+  pytest
+  pytest-cov
+  pytest-operator
```
samuelallan72, line 83: note: ideally projects would put all the test dependencies under tests/*/requirements.txt, so we wouldn't need extras here. But most projects have pytest* here, so I left it for now.

chanchiwai-ray (approved):
We can still fix any issue on the PR created

chanchiwai-ray (approved):

chanchiwai-ray (dismissed):

chanchiwai-ray (approved):

terraform-plans/templates/github/tox.ini.tftpl
```diff
@@ -0,0 +1,103 @@
...
+  -r {toxinidir}/requirements.txt
+  -r {toxinidir}/tests/functional/requirements.txt
+%{ if functest_type == "zaza" ~}
+changedir = {toxinidir}/tests/functional
+commands = functest-run-suite {posargs:--keep-model -v}
+%{ else ~}%{ if functest_type == "pytest" ~}
+commands = pytest {toxinidir}/tests/functional {posargs:-v}
```
gabrielcocenza, line 98: To say the truth, I don't know if we have a project that generates coverage for functional tests. However, I think we can start be ready for that. E.g: COU we can possibly use the functional tests with the python package instead of the snap to have some coverage.

With that said, I think we can use the same arguments of unit test to generate coverage.

gabrielcocenza (changes requested):

terraform-plans/templates/github/tox.ini.tftpl
```diff
@@ -0,0 +1,103 @@
...
+commands = echo "No unit tests, skipping."
+%{ else ~}
+%{ if unittest_type == "pytest" ~}
+commands = pytest {toxinidir}/tests/unit \
+   -v \
+   --cov \
+   --cov-report=term-missing \
+   --cov-report=html \
+   --cov-report=xml \
+   {posargs}
```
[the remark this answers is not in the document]
samuelallan72: What is your use case for running without coverage? The use cases I have in mind are:

- `tox -e unit` -> runs same as in CI - all the tests, with coverage and reports
- `tox -e unit -- <mytest>` -> runs same as in CI with coverage, but only runs a specific test or set of tests.

In your example, you can still add `--disable-warnings` with how it is currently.

terraform-plans/templates/github/tox.ini.tftpl
samuelallan72: Ah agreed, that would be good. And we can also get coverage from the test .py files themselves, which is also important. :+1:

terraform-plans/templates/github/tox.ini.tftpl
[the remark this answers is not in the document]
gabrielcocenza: I don't have a good use case in the top of my head a situation that is very important to not generate the coverage.

Maybe when trying to run just a specific test like with more verbose:

```
tox -e unit -- -k "test_foo" -vvvv
```

On your solution the `-vvvv` would append at the end and there is already a `-v`. Not sure if it's problematic, but I think it's a nice standard of having what we expect for the CI to run with pure `tox -e unit` and be responsible to add all arguments if you want to try something different.

I won't block because of this, so feel free to resolve

gabrielcocenza (approved):

chanchiwai-ray (approved):

PR #161, opened 2025-03-03
Update check and release workflow to release only tested charms

chanchiwai-ray:
Changes in this PR need to be applied manually for these projects since they only have `check.yaml` or `release.yaml`.

1. charm-prometheus-juju-exporter
2. charm-simple-streams                                                                                                                                                                                                                                                                      
3. hardware-observer-operator


samuelallan72 (approved):
LGTM, thanks!

Pjack (approved):

terraform-plans/templates/github/charm_check.yaml.tftpl
```diff
@@ -72,21 +72,53 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - uses: canonical/craft-actions/charmcraft/pack@main
+        id: build
+        with:
+          verbosity: verbose
+          channel: "${ charmcraft_channel }"
```
samuelallan72, line 95: Can we check the details of how this caches things here? I have some doubts:

- if there are multiple jobs, they will be likely running in parallel, so they won't be able to benefit from the cache of the other
- we're only going to be running multiple jobs if we need to build on different architectures. I'm worried that cache artefacts from one architecture will be different to that of another architecture.

Maybe the cache is a good idea; I haven't looked far into the details. But my personal thought on this is I'd rather ensure the charm is built from a true clean reproducible environment, because this will be the production charm we're publishing. If the build takes a bit longer, it's still worth it imo.

terraform-plans/templates/github/charm_release.yaml.tftpl
```diff
@@ -19,25 +19,25 @@ jobs:
...
 
       - name: Pack and upload to charmhub
-        uses: canonical/charming-actions/upload-charm@2.6.2
+        uses: canonical/charming-actions/upload-charm@2.7.0
         with:
-          charmcraft-channel: "${charmcraft_channel}"
           credentials: "$${{ secrets.CHARMHUB_TOKEN }}"
           github-token: "$${{ secrets.GITHUB_TOKEN }}"
-          # Ensure the charm is built in an isolated environment and on the correct base in an lxd container.
-          destructive-mode: false
```
Pjack, line 43: It's not available in the new build procedure anymore. Why don't we need it anymore?

chanchiwai-ray: because we don't build anymore

terraform-plans/templates/github/charm_check.yaml.tftpl
chanchiwai-ray: it's based on the following https://github.com/canonical/craft-actions/blob/main/charmcraft/pack/action.yaml#L58

chanchiwai-ray: I did not think about the multi-arch scenario. I was thinking about e.g. rebuilding charm during a PR, where you can constantly pushing different commit without changing the dependency

terraform-plans/templates/github/charm_release.yaml.tftpl
Pjack: We have `build` process in  [charm_check.yaml.tftpl](https://github.com/canonical/solutions-engineering-automation/pull/161/files#diff-97ae0cd1ad9d06cfe014821942219ccf29e0797bdc5c3c3cd3067cfa6372499d). Should we disable/enable destructive-mode ?

chanchiwai-ray: `charmcraft pack` by default runs in **non** destructive mode

terraform-plans/templates/github/charm_check.yaml.tftpl
samuelallan72: Ah of course, I wasn't sure how long the caches lived - that could definitely be helpful if it works across test runs. Maybe we need to look into setting a custom cache key if we're going to use the cache? To ensure that it depends on the architecture (if that applies) and any requirements files, etc. we use that might not be picked up by the default cache key.

Although even then, if we want to be extra careful, dependencies can change even if our requirements.txt doesn't change, because we're not using lock files. It feels a little risky to me if we're using this for the production charm uploads. :sweat_smile:

samuelallan72: Or maybe something like only using the cache for code running in PRs, or PRs in draft mode when there may be commits pushed frequently? That might help for saving time there, but ensuring a clean environment for the production build?

chanchiwai-ray: Yeah, thanks for that, maybe we should just rollback we previously did for now. There's still at least on unclear points from me based on your question: does this work properly in multi-arch case

terraform-plans/templates/github/charm_release.yaml.tftpl
Pjack: Thanks! I find the document in 
https://github.com/canonical/craft-actions?tab=readme-ov-file#charmcraftpack

terraform-plans/templates/github/charm_check.yaml.tftpl
chanchiwai-ray: btw charmcraft team recommend using this https://canonical-charmcraft.readthedocs-hosted.com/en/stable/howto/build-guides/shared-cache/#on-github

Pjack (approved):

terraform-plans/configs/charm-sysconfig_main.tfvars
```diff
@@ -18,8 +18,8 @@ templates = {
     vars = {
       # Skip ARM64 check because the functional test runs on lxd VM which is not working
```
gabrielcocenza, line 19: I guess now we can remove this comment because we are adding arm64?

chanchiwai-ray: Removed

gabrielcocenza (approved):

samuelallan72 (approved):
LGTM :+1:

PR #163, opened 2025-03-04
feat: change check default runs-on to noble and support noble for charm-local-user

jneo8:
- Support noble for charm-local-user
- Change lint and unit default runs-on to noble 

See https://github.com/canonical/charm-local-users/pull/62 for required change.


chanchiwai-ray (commented):
I think you can just change the default value for `python-version` in the template since func test is not dependent on that most likely

jneo8:
> I think you can just change the default value for `python-version` in the template since func test is not dependent on that most likely

Since 3.12 will only be used by some of the project, I try not to affect the projects not relate to it.

chanchiwai-ray:
I think the risk is low to change the template directly. Besides, it seems you forgot to update the `runs-on` in `lint` and `unit` job

chanchiwai-ray (approved):
👍🏼

Pjack (approved):

PR #166, opened 2025-03-10
Standardize Allowed GitHub Actions on SolEng projects

gabrielcocenza:
Some projects have full permissions to allow all actions and reusable workflows that can be a security issue.

With this PR, all projects will run actions or reusable workflows that are created by GitHub or with the following pattern:
- canonical/*
- snapcore/*
- charmed-kubernetes/*
- tiobe/*
- beliaev-maksim/github-to-jira-automation/.github/workflows/issues_to_jira.yaml@master

See more information about how to configure those options in the terraform [documentation](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/actions_repository_permissions)

With those changes the idea is to have repos configured like this:

image

### DIFF
--- a/terraform-plans/modules/GitHub/settings/main.tf
+++ b/terraform-plans/modules/GitHub/settings/main.tf
@@ -32,6 +32,25 @@ resource "github_repository" "repo" {
 
 }
 
+resource "github_actions_repository_permissions" "repo" {
+  repository      =  var.repository
+  allowed_actions = "selected"
+
+  allowed_actions_config {
+    github_owned_allowed = true
+
+    verified_allowed = false
+
+    patterns_allowed = [
+      beliaev-maksim/github-to-jira-automation/.github/workflows/issues_to_jira.yaml@master,
+      canonical/*,
+      snapcore/*,
+      charmed-kubernetes/*,
+      tiobe/*,
+    ]
+  }
+}
+
 data "github_team" "admins" {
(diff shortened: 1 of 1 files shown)


terraform-plans/modules/GitHub/settings/main.tf
```diff
@@ -33,6 +33,25 @@ resource "github_repository" "repo" {
...
+
+  allowed_actions_config {
+    github_owned_allowed = true
+
+    verified_allowed = false
+
+    patterns_allowed = [
+      "beliaev-maksim/github-to-jira-automation/.github/workflows/issues_to_jira.yaml@master",
+      "canonical/*",
+      "snapcore/*",
```
samuelallan72, line 48: All repositories in https://github.com/snapcore have an equivalent in github.com/canonical/ - can we find out from the snap / store team which are the canonical sources?

gabrielcocenza: Sorry, I think I didn't understand your question...

So based on what you said, we can remove the `snapcore/*`?

samuelallan72: I think so, but some of our projects still use the `snapcore/*` versions. We could check and standardise. For now this is probably fine.

samuelallan72 (approved):

chanchiwai-ray (approved):

PR #177, opened 2025-03-14
Add extra_steps that is need for tests in COU and juju-backup-all

gabrielcocenza:
COU and juju-backup-all has a special case that an extra step is needed to run the functional tess.

This change adds the `needs_juju` that can be added to add the necessary extra step.

The intention is trying to keep a single template for all projects, including COU and this will make easier for future changes or integrations.

See:
Expected [COU workflow](https://github.com/canonical/solutions-engineering-automation/actions/runs/13907380260/job/38913506913?pr=177#step:6:322)
[Example](https://github.com/canonical/solutions-engineering-automation/actions/runs/13907380260/job/38913516455?pr=177#step:6:327) of workflow of project that is not necessary an extra step

### DIFF
--- a/terraform-plans/templates/github/snap_check.yaml.tftpl
+++ b/terraform-plans/templates/github/snap_check.yaml.tftpl
@@ -135,8 +135,8 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.10"
-%{ if extra_steps != "" }
-${indent(6, chomp(extra_steps))}
+%{ if extra_func_steps != "" }
+${indent(6, chomp(extra_func_steps))}
 %{ endif }
       - name: Install dependencies
         run: |
--- a/terraform-plans/templates/github/snap_check.yaml.tftpl
+++ b/terraform-plans/templates/github/snap_check.yaml.tftpl
@@ -137,7 +137,7 @@ jobs:
           python-version: "3.10"
 
 %{ if needs_juju == "true" ~}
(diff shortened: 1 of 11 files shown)


terraform-plans/templates/github/snap_check.yaml.tftpl
```diff
@@ -135,7 +135,9 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.10"
-
+%{ if extra_func_steps != "" }
+${indent(6, chomp(extra_func_steps))}
+%{ endif }
```
samuelallan72, line 140: I'm really not a fan of templating yaml within yaml... it's very prone to error.

Ideally we're wanting to keep the templates as consistent as possible, so having a freeform 'extra steps' field goes against that ideal. I'd like to avoid too many variables for the templates, especially variables that have large values.

How about a variable like `needs_juju = "true"`? Or just always installing juju? Or perhaps a separate template for these two projects that need juju?

gabrielcocenza: I see your point. I would like to avoid having two separate templates, because it's very easy to forget to sync the common parts if changes are needed.

I'll apply your suggestion of using `needs_juju`

samuelallan72 (approved):
LGTM :)

Deezzir (approved):
LGTM

PR #178, opened 2025-03-15
Fix the logic when there is no coverage files

gabrielcocenza:
- the logic when there is no coverage files were failing because `coverage_files=./artifacts/.coverage*` always contains the wildcard as a string rather than expanding to actual files.
- include the missing env.TEST_CMD_ID for charms because functional tests can have multiple commands and artifacts cannot have the same name

A POC that this change is working can be found at https://github.com/canonical/charm-prometheus-blackbox-exporter/pull/34 that was one of the projects that were failing because of this

### DIFF
--- a/terraform-plans/templates/github/charm_check.yaml.tftpl
+++ b/terraform-plans/templates/github/charm_check.yaml.tftpl
@@ -275,9 +275,9 @@ jobs:
 
           coverage_files=(./artifacts/.coverage*)
 
-          if [ -e "${coverage_files[0]}" ]; then
-            echo "Merging coverage files: ${coverage_files[*]}"
-            coverage combine "${coverage_files[@]}"
+          if [ -e "$${coverage_files[0]}" ]; then
+            echo "Merging coverage files: $${coverage_files[*]}"
+            coverage combine "$${coverage_files[@]}"
             coverage xml -o tests/report/coverage.xml
           else
             echo "No coverage files found, skipping merge"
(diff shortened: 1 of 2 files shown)


terraform-plans/templates/github/charm_check.yaml.tftpl
```diff
@@ -275,9 +275,9 @@ jobs:
 
           coverage_files=./artifacts/.coverage*
```
samuelallan72, line 276: This also needs parens to match the snap_check.yaml template changes

```suggestion
          coverage_files=(./artifacts/.coverage*)
```

samuelallan72 (changes requested):
Nice catches! I just flagged one place where it didn't match between the two templates.

terraform-plans/templates/github/charm_check.yaml.tftpl
```diff
@@ -202,7 +202,7 @@ jobs:
         with:
           include-hidden-files: true
           if-no-files-found: ignore
-          name: coverage-functional-$${{ env.SYSTEM_ARCH }}
+          name: coverage-functional-${{ env.TEST_CMD_ID }}-${{ env.SYSTEM_ARCH }}
```
samuelallan72, line 205: The `$`'s need escaping here:

```suggestion
          name: coverage-functional-$${{ env.TEST_CMD_ID }}-$${{ env.SYSTEM_ARCH }}
```

samuelallan72 (approved):

Vultaire (approved):
LGTM

PR #179, opened 2025-03-18
Add missing projects to the weekly run

gabrielcocenza:
Some projects were missing from our weekly run which can hide problems if the CI is broken

### DIFF
--- a/.github/workflows/weekly_tests.yaml
+++ b/.github/workflows/weekly_tests.yaml
@@ -33,13 +33,17 @@ jobs:
           - charm-storage-connector
           - charm-sysconfig
           - charm-userdir-ldap
+          - charmed-openstack-exporter-snap
           - charmed-openstack-upgrader
+          - dcgm-snap
+          - hardware-observer-operator
           - juju-backup-all
           - juju-lint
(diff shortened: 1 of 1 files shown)


samuelallan72 (approved):

.github/workflows/weekly_tests.yaml
```diff
@@ -33,13 +33,17 @@ jobs:
           - charm-storage-connector
           - charm-sysconfig
           - charm-userdir-ldap
+          - charmed-openstack-exporter-snap
           - charmed-openstack-upgrader
+          - dcgm-snap
+          - hardware-observer-operator
```
samuelallan72, line 39: hardware-observer-operator is already in the matrix here - see the include clause below. :)

samuelallan72 (changes requested):

.github/workflows/weekly_tests.yaml
gabrielcocenza: Thanks for catching that

samuelallan72 (approved):

Deezzir (approved):

PR #180, opened 2025-03-19
Temporarily disable mypy in tox.ini

samuelallan72:
It doesn't play nice if there are no py files find at the path, so we can't hardcode this for all repos.
We need to figure out a solution, but I don't think it's high priority, so let's disable it temporarily now to unblock things.

### DIFF
--- a/terraform-plans/templates/github/tox.ini.tftpl
+++ b/terraform-plans/templates/github/tox.ini.tftpl
@@ -23,7 +23,8 @@ commands =
     isort --check --diff --color .
     flake8
     pylint --recursive=y .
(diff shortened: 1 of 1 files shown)


jneo8 (approved):
It's a temporary workaround to avoid blocking other PRs in downstream. But we do need to check the effort we need to have common tox.ini for every project to decide next step.

@samuelallan72 can you help to open an issue to trace this one? I am thinking we can have a check list for all the projects there and everyone can help to check part of them.

samuelallan72:
@jneo8 a spreadsheet has been created and shared internally. :)

Pjack (approved):

PR #186, opened 2025-03-20
Add release into missing projects

gabrielcocenza:
Some projects were missing the release template.
Others doesn't need but it's good to have a comment explaining why.


jneo8 (approved):
Nice one!

chanchiwai-ray (approved):

PR #188, opened 2025-03-21
fix: Options to have requirements.txt as dep in tox.ini

jneo8:
Provide options to include requirements.txt as dependency in tox.ini.

- root_requirements_txt: include in project's `./`
- unit_requirements_txt: include in prject's `./tests/unit`
- func_requirements_txt: include in prject's `./tests/func`
- Ignore dep when test type is none in lint testenv.

### DIFF
--- a/terraform-plans/configs/charm-advanced-routing_main.tfvars
+++ b/terraform-plans/configs/charm-advanced-routing_main.tfvars
@@ -69,8 +69,11 @@ templates = {
     source      = "./templates/github/tox.ini.tftpl"
     destination = "tox.ini"
     vars = {
-      functest_type = "pytest"
-      unittest_type = "pytest"
+      functest_type         = "pytest"
+      unittest_type         = "pytest"
+      root_requirements_txt = "yes"
+      unit_requirements_txt = "yes"
+      func_requirements_txt = "yes"
     }
   }
 }
--- a/terraform-plans/templates/github/tox.ini.tftpl
+++ b/terraform-plans/templates/github/tox.ini.tftpl
@@ -38,8 +38,12 @@ deps =
     pep8-naming
     pylint
     # so pylint and mypy can reason about the code
+%{ if unittest_type != "none" ~}
     {[testenv:unit]deps}
+%{ endif ~}
+%{ if functest_type != "none" ~}
     {[testenv:func]deps}
+%{ endif ~}
 
 [testenv:reformat]
 commands =
@@ -70,8 +74,12 @@ commands = pytest {toxinidir}/tests/unit \
 deps =
   pytest
   pytest-cov
+%{ if root_requirements_txt == "yes" ~}
   -r {toxinidir}/requirements.txt
+%{ endif ~}
+%{ if unittest_type != "none" && unit_requirements_txt == "yes" ~}
   -r {toxinidir}/tests/unit/requirements.txt
+%{ endif ~}
 %{ else ~}
 ERROR: invalid `unittest_type` value
 %{ endif ~}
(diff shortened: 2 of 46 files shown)


chanchiwai-ray (approved):

gabrielcocenza (commented):
Is an issue to go on repos that is giving this issue and create empty requirements files?

I think it would be easier to maintain and introduce less logic in the templates and need to map more things in the projects

terraform-plans/configs/charm-advanced-routing_main.tfvars
```diff
@@ -69,8 +69,11 @@ templates = {
     source      = "./templates/github/tox.ini.tftpl"
     destination = "tox.ini"
     vars = {
-      functest_type = "pytest"
-      unittest_type = "pytest"
+      functest_type         = "pytest"
+      unittest_type         = "pytest"
+      root_requirements_txt = "yes"
+      unit_requirements_txt = "yes"
+      func_requirements_txt = "yes"
```
samuelallan72, line 76: IMHO this goes over the line, introducing too much complexity to the template - three more low level variables. Can we find a better way please?

terraform-plans/templates/github/tox.ini.tftpl
```diff
@@ -38,8 +38,12 @@ deps =
     pep8-naming
     pylint
     # so pylint and mypy can reason about the code
+%{ if unittest_type != "none" ~}
     {[testenv:unit]deps}
+%{ endif ~}
+%{ if functest_type != "none" ~}
     {[testenv:func]deps}
+%{ endif ~}
```
samuelallan72, line 46: This part makes sense though.

terraform-plans/templates/github/tox.ini.tftpl
```diff
@@ -70,8 +74,12 @@ commands = pytest {toxinidir}/tests/unit \
 deps =
   pytest
   pytest-cov
+%{ if root_requirements_txt == "yes" ~}
   -r {toxinidir}/requirements.txt
+%{ endif ~}
+%{ if unittest_type != "none" && unit_requirements_txt == "yes" ~}
   -r {toxinidir}/tests/unit/requirements.txt
+%{ endif ~}
```
samuelallan72, line 82: If there are python based unit tests, could we simply make it a convention that we always include a requirements.txt file in the project? Even if it's empty.

samuelallan72 (changes requested):
I think this introduces too much low level variability to the template.

terraform-plans/templates/github/tox.ini.tftpl
samuelallan72: Actually, I don't think this is needed.

terraform-plans/configs/charm-advanced-routing_main.tfvars
samuelallan72: As discussed in the meeting, we can use a single new variable: `is_python_project`.

terraform-plans/templates/github/tox.ini.tftpl
```diff
@@ -90,8 +98,12 @@ deps =
   pytest
   pytest-cov
   pytest-operator
+%{ if root_requirements_txt == "yes" ~}
   -r {toxinidir}/requirements.txt
+%{ endif ~}
```
samuelallan72, line 103:
```suggestion
%{ if is_python_project == "yes" ~}
  -r {toxinidir}/requirements.txt
%{ endif ~}
```

terraform-plans/templates/github/tox.ini.tftpl
```diff
@@ -90,8 +98,12 @@ deps =
...
+%{ if functest_type != "none" && func_requirements_txt == "yes" ~}
   -r {toxinidir}/tests/functional/requirements.txt
+%{ endif ~}
```
samuelallan72, line 106: This can be dropped

terraform-plans/templates/github/tox.ini.tftpl
samuelallan72: Either way, I'm not fussed about this.

samuelallan72:
Superceded by #189

PR #189, opened 2025-03-25
fix: linter & dependency issues on tox.ini template

jneo8:
Fix couple issues on tox.ini:

- Options to disable pylint & mypy on tox.ini, default disable.
- Option to include requirements.txt in root directory.
- Include testenv deps only if test is not none.

### DIFF
--- a/terraform-plans/templates/github/tox.ini.tftpl
+++ b/terraform-plans/templates/github/tox.ini.tftpl
@@ -22,10 +22,10 @@ commands =
     black --check --diff --color .
     isort --check --diff --color .
     flake8
-%{ if enable_pylint == "yes" ~}
+%{ if enable_pylint == "true" ~}
     pylint --recursive=y .
 %{ endif ~}
-%{ if enable_mypy == "yes" ~}
+%{ if enable_mypy == "true" ~}
     mypy --install-types --non-interactive .
 %{ endif ~}
 deps =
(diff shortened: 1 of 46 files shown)


chanchiwai-ray (approved):
I checked action output for snap-tempest and logroated, and they seems okay

terraform-plans/templates/github/tox.ini.tftpl
```diff
@@ -22,9 +22,12 @@ commands =
     black --check --diff --color .
     isort --check --diff --color .
     flake8
+%{ if enable_pylint == "yes" ~}
     pylint --recursive=y .
-    # mypy temporarily disabled because it doesn't gracefully recursively find files like the others
-    #mypy --install-types --non-interactive .
+%{ endif ~}
+%{ if enable_mypy == "yes" ~}
+    mypy --install-types --non-interactive .
```
samuelallan72, line 29: It's checking for "yes" here, but in the variables it's setting "true" or "false"

samuelallan72 (changes requested):

terraform-plans/templates/github/tox.ini.tftpl
jneo8: Ah, I forget change it to true. Thanks for find it out.

samuelallan72 (approved):

chanchiwai-ray (approved):

PR #191, opened 2025-03-25
functest is zaza need TEST_JUJU3 env

chanchiwai-ray:
See https://github.com/canonical/charmed-openstack-upgrader/actions/runs/14059451465/job/39366428374.

Change is applied to COU, this only ports the change back to automation

### DIFF
--- a/terraform-plans/templates/github/tox.ini.tftpl
+++ b/terraform-plans/templates/github/tox.ini.tftpl
@@ -15,6 +15,9 @@ setenv =
   PYTHONPATH = {toxinidir}:{toxinidir}/src/:{toxinidir}/reactive/:{toxinidir}/hooks/:{toxinidir}/lib/:{toxinidir}/actions:{toxinidir}/files/:{toxinidir}/files/plugins/
   # avoid state written to file during tests - see https://github.com/juju/charm-helpers/blob/85dcbeaf63b0d0f38e8cb17825985460dc2cd02d/charmhelpers/core/unitdata.py#L179-L184
   UNIT_STATE_DB = :memory:
+%{ if functest_type == "zaza" ~}
+  TEST_JUJU3 = 1
+%{ endif ~}
 passenv = *
 
 [testenv:lint]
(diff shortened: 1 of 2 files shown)


terraform-plans/templates/github/tox.ini.tftpl
```diff
@@ -15,6 +15,9 @@ setenv =
   PYTHONPATH = {toxinidir}:{toxinidir}/src/:{toxinidir}/reactive/:{toxinidir}/hooks/:{toxinidir}/lib/:{toxinidir}/actions:{toxinidir}/files/:{toxinidir}/files/plugins/
   # avoid state written to file during tests - see https://github.com/juju/charm-helpers/blob/85dcbeaf63b0d0f38e8cb17825985460dc2cd02d/charmhelpers/core/unitdata.py#L179-L184
   UNIT_STATE_DB = :memory:
+%{ if functest_type == "zaza" ~}
+  TEST_JUJU3 = 1
+%{ endif ~}
```
samuelallan72, line 20: This can be unconditionally set.

```suggestion
  # So zaza tests work with python 3 https://github.com/openstack-charmers/zaza/pull/653
  TEST_JUJU3 = 1
```

Note that also the charm check template includes this env var: https://github.com/canonical/solutions-engineering-automation/blob/main/terraform-plans/templates/github/charm_check.yaml.tftpl#L192

It does make more sense to put it directly in tox.ini, to facilitate running tests locally.

Please confirm you've tested this - will this env be visible to the process when tox is installing the dependencies?

chanchiwai-ray: please check https://github.com/canonical/charmed-openstack-upgrader/pull/617, adding is to tox env makes the lint pass

chanchiwai-ray: before there will be conflict between juju and zaza

samuelallan72: OK thanks for checking.

To clean up:

- TEST_JUJU3=1 can be unconditionally set
- TEST_JUJU3 should be removed from the check.yaml if we're adding it to tox.ini.

chanchiwai-ray: done

samuelallan72 (approved):

Deezzir (approved):

Issue #105, opened 2024-10-08
Including commit message or PR number in automated PR description


rgildein:
It would be great if we can track original PR from automated PR. For example, [this](https://github.com/canonical/charm-advanced-routing/pull/57) could provide some information, maybe in PR description, about the original PR in this repo. I think it was this PR #103.

In all automated PR we can see only this:
> This is an automated pull request from https://github.com/canonical/solutions-engineering-automation to update centrally managed files.

However, It would be great to see original PR with some context why and what we are changing. Maybe see any comments.

PR #196, opened 2025-03-31
feat: add additional description to PRs created with terraform

Deezzir:
This adds a Python script that runs after the `Terraform apply` step. The script will append the created PR with the PR link from this repo, adding a changelog. 

Having a script may allow us to do other things to improve the automation, giving us more flexibility. 

We can also stick to the Terraform plan, but due to limitations, we won't be able to have a `Changelog`, which accumulates a list of the PRs on the target repos.

[Example PR description](https://github.com/Deezzir/solutions-engineering-automation/pull/1)

Closes: #105

### DIFF
--- a/.github/workflows/terraform-apply.yaml
+++ b/.github/workflows/terraform-apply.yaml
@@ -130,11 +130,20 @@ jobs:
         run: |
           terraform apply ./tf.plan -no-color
 
+      - name: Create GitHub App Token
+        if: ${{ github.ref_name == 'main' && github.event_name != 'workflow_dispatch' }}
+        uses: actions/create-github-app-token@v1
+        id: app_token
+        with:
+          app_id: ${{ secrets.SOLENG_APP_ID }}
+          installation_id: ${{ secrets.SOLENG_APP_INSTALLATION_ID }}
+          private_key: ${{ secrets.SOLENG_APP_PEM_FILE }}
+
       - name: Update PR description with commit info (if running from main branch)
         if: ${{ github.ref_name == 'main' && github.event_name != 'workflow_dispatch' }}
         working-directory: ./scripts
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ steps.app_token.outputs.token }}
         run: |
           sudo apt-get update
           sudo apt-get install -y python3 python3-pip
--- a/.github/workflows/terraform-apply.yaml
+++ b/.github/workflows/terraform-apply.yaml
@@ -129,3 +129,34 @@ jobs:
           GITHUB_APP_PEM_FILE: ${{ secrets.SOLENG_APP_PEM_FILE }}
         run: |
           terraform apply ./tf.plan -no-color
+
+      - name: Update PR description with commit info
+        if: ${{ github.ref_name == 'main' }}
+        working-directory: ./scripts
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y python3 python3-pip
+          pip3 install requests
+
+          COMMIT_SHA="${{ github.sha }}"
+
+          PR_URL=$(terraform output -raw pr_url)
+          PR_CREATED=$(terraform output -raw pr_created)
+
+          echo "Commit SHA: $COMMIT_SHA"
+          echo "PR URL: $PR_URL"
+          echo "PR Created: $PR_CREATED"
+
+          if [[ "$PR_URL" == "No PR created" ]]; then
+            echo "No PR was created — skipping PR update step."
+            exit 0
+          fi
+
+          SCRIPT_ARGS="--url $PR_URL --commit $COMMIT_SHA --token $GITHUB_TOKEN"
+          if [[ "$PR_CREATED" == "false" ]]; then
+            SCRIPT_ARGS="$SCRIPT_ARGS -a"
+          fi
+
+          python3 update-pr-desc.py $SCRIPT_ARGS
--- a/.github/workflows/terraform-apply.yaml
+++ b/.github/workflows/terraform-apply.yaml
@@ -130,7 +130,7 @@ jobs:
         run: |
           terraform apply ./tf.plan -no-color
 
-      - name: Update PR description with commit info
+      - name: Update PR description with commit info (if running from main branch)
         if: ${{ github.ref_name == 'main' }}
         working-directory: ./scripts
         env:
--- a/.github/workflows/terraform-apply.yaml
+++ b/.github/workflows/terraform-apply.yaml
@@ -141,7 +141,6 @@ jobs:
           pip3 install requests
 
           COMMIT_SHA="${{ github.sha }}"
-
           PR_URL=$(terraform output -raw pr_url)
           PR_CREATED=$(terraform output -raw pr_created)
 
@@ -156,7 +155,7 @@ jobs:
 
           SCRIPT_ARGS="--url $PR_URL --commit $COMMIT_SHA --token $GITHUB_TOKEN"
           if [[ "$PR_CREATED" == "false" ]]; then
-            SCRIPT_ARGS="$SCRIPT_ARGS -a"
+            SCRIPT_ARGS="$SCRIPT_ARGS --append"
           fi
 
           python3 update-pr-desc.py $SCRIPT_ARGS
--- a/.github/workflows/terraform-apply.yaml
+++ b/.github/workflows/terraform-apply.yaml
@@ -131,7 +131,7 @@ jobs:
           terraform apply ./tf.plan -no-color
 
       - name: Update PR description with commit info (if running from main branch)
-        if: ${{ github.ref_name == 'main' }}
+        if: ${{ github.ref_name == 'main' && github.event_name != 'workflow_dispatch' }}
         working-directory: ./scripts
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


Deezzir:
Is there a way to test this? I've tested the script a PR created for one of my forks

.github/workflows/terraform-apply.yaml
```diff
@@ -129,3 +129,33 @@ jobs:
...
+          echo "PR Created: $PR_CREATED"
+
+          if [[ "$PR_URL" == "No PR created" ]]; then
+            echo "No PR was created — skipping PR update step."
+            exit 0
+          fi
+
+          SCRIPT_ARGS="--url $PR_URL --commit $COMMIT_SHA --token $GITHUB_TOKEN"
+          if [[ "$PR_CREATED" == "false" ]]; then
+            SCRIPT_ARGS="$SCRIPT_ARGS --append"
```
samuelallan72, line 158: This workflow could be manually run an arbitrary number of times from main - so this could append duplicate entries

.github/workflows/terraform-apply.yaml
```diff
@@ -129,3 +129,33 @@ jobs:
...
+          echo "Commit SHA: $COMMIT_SHA"
+          echo "PR URL: $PR_URL"
+          echo "PR Created: $PR_CREATED"
+
+          if [[ "$PR_URL" == "No PR created" ]]; then
+            echo "No PR was created — skipping PR update step."
+            exit 0
+          fi
+
+          SCRIPT_ARGS="--url $PR_URL --commit $COMMIT_SHA --token $GITHUB_TOKEN"
```
samuelallan72, line 156: It's a little dangerous to put secrets on the cli. Maybe the script could read it from the env?

samuelallan72 (commented):
> Is there a way to test the CI?

Outside of creating a fork of the automation repo in a separate org and configuring it with a temporary repository to control, not really... :(

---

This is a pretty cool change - please see some thoughts inline too.

---

Personally though I don't have any issues with the way things are currently - the automated PRs are simply getting the target repos inline with the automation repo. If the target repos were changed to be out of line, then the automated PR will bring it into line, regardless of which recent PR was merged to automation. As in, linking to commits from automation in this way can only give a best-effort overview.

The real source of truth can be obtained by checking the corresponding template in automation, and running a `git blame` on it.

.github/workflows/terraform-apply.yaml
Deezzir: Addressed in 7e19a508e1177703cb0a2bddf1f2fc8cd0ceec73

.github/workflows/terraform-apply.yaml
Deezzir: Addressed in 7e19a508e1177703cb0a2bddf1f2fc8cd0ceec73

Deezzir:
@samuelallan72 Thanks for the thoughts you provided. The current setup will only append the PRs if something was merged to the `solutions-engineering-automation` repo. If the workflow is run manually(without a commit SHA), the description will stay as it is. The goal is to provide some context on what is happening in case of a template change.

We can discuss it further

scripts/update-pr-desc.py
```diff
@@ -0,0 +1,169 @@
...
+    parser.add_argument("--url", required=True, help="URL of the PR to be updated")
+    parser.add_argument(
+        "--commit",
+        required=True,
+        help="The Commit SHA which will be included in the PR's description",
+    )
+    parser.add_argument(
+        "--token",
+        help="Github Token to use for the authorization (can also use GITHUB_TOKEN env var)",
+    )
```
samuelallan72, line 148: We can be completely opinionated here - nothing is using --token now, so it can be removed. :)

Deezzir: Removed

jneo8 (dismissed):
Very nice improvement.

Can you also add the example output in the pr description?

scripts/update-pr-desc.py
```diff
@@ -0,0 +1,165 @@
...
+        help="Whether to append to the existing list of commits or create it",
+        action="store_true",
+    )
+
+    args = parser.parse_args()
+    token = os.getenv("GITHUB_TOKEN")
+
+    if not token:
+        parser.error(
+            "Github token is required. Provide it with --token or set GITHUB_TOKEN environment variable."
```
samuelallan72, line 158:
```suggestion
            "Github token is required. Provide it by setting the GITHUB_TOKEN environment variable."
```

I think this was missed when removing `--token` support

chanchiwai-ray (approved):
LGTM, the example looks nice 👍🏼

scripts/update-pr-desc.py
```diff
@@ -0,0 +1,165 @@
...
+    if commit_url in body:
+        logger.warning("Commit already in the PR's description")
+        return body
+
+    changelog_header = ""
+    if not append and CHANGELOG_HEADER not in body:
+        changelog_header = f"\n\n{CHANGELOG_HEADER}\n"
+    new_body = f"{body.rstrip()}{changelog_header}\n- {commit_url}"
+
+    return new_body
```
gabrielcocenza, line 112: **suggestion**

I'm favorable of letting terraform it self being responsible for the first message and this script literally just update and not create from scratch.

I think we can achieve this by declaring a new variable in terraform like:
```
variable "commit_sha" {
  description = "The commit SHA of the PR"
  type        = string
}
```

And then pass in the workflow like:

```
terraform apply ./tf.plan -no-color -var="commit_sha=${{ github.sha }}
```

I think this reduces the work in the script side and it will run only if there is a PR_URL and no `PR_CREATED`.

What do you think?

This means less logic here, no need of the append argument

gabrielcocenza (changes requested):
There is just one blocker change and I left some comments that I think it would be a nice improvement, but not needed.

scripts/update-pr-desc.py
Deezzir: I did it with the script, i.e, appending the `CHANGELOG` header, because the workflow can be triggered manually and `${{ github.sha }}` will be empty. The step that runs the script checks that and only runs on merge to `main`. I guess we can check that will terraform as well, but my goal was to leave the terraform logic intact and make the script more of an addition to the existing workflow that can be removed/added if we don't like it without touching the main automation logic.

.github/workflows/terraform-apply.yaml
```diff
@@ -129,3 +129,33 @@ jobs:
...
+
+      - name: Update PR description with commit info (if running from main branch)
+        if: ${{ github.ref_name == 'main' && github.event_name != 'workflow_dispatch' }}
+        working-directory: ./scripts
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y python3 python3-pip
+          pip3 install requests
```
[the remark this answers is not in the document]
Deezzir: Leaving the setup in the same step is okay, so we don't waste resources on each merge or manual dispatch for the installation. In the current implementation, I safeguard by checking:

```
if: ${{ github.ref_name == 'main' && github.event_name != 'workflow_dispatch' }}
``

So the installation only occurs if we need it. What do you think?

File: scripts/update-pr-desc.py [hunk 7]
```diff
@@ -0,0 +1,165 @@
...
+        },
+        json={"body": body},
+    )
+
+    if resp.status_code == 403 or resp.status_code == 401:
+        raise PermissionError("Request to update the PR's information was unauthorized")
+    if resp.status_code == 404:
+        raise ValueError("PR not found")
+    if resp.status_code != 200:
+        raise ValueError(f"Failed to update the PR: {resp.json()}")
```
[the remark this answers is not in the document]
Deezzir: Addressed in e6335a85d00a9a5ce89a9a6947ad66079b6401b4

scripts/update-pr-desc.py
```diff
@@ -0,0 +1,165 @@
...
+
+    args = parser.parse_args()
+    token = os.getenv("GITHUB_TOKEN")
+
+    if not token:
+        parser.error(
+            "Github token is required. Provide it by setting the GITHUB_TOKEN environment variable."
+        )
+
+    update_description(args.url, args.commit, args.token, args.append)
```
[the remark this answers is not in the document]
Deezzir: Addressed in e6335a85d00a9a5ce89a9a6947ad66079b6401b4

.github/workflows/terraform-apply.yaml
```diff
@@ -129,3 +129,33 @@ jobs:
...
+          echo "PR URL: $PR_URL"
+          echo "PR Created: $PR_CREATED"
+
+          if [[ "$PR_URL" == "No PR created" ]]; then
+            echo "No PR was created — skipping PR update step."
+            exit 0
+          fi
+
+          SCRIPT_ARGS="--url $PR_URL --commit $COMMIT_SHA"
+          if [[ "$PR_CREATED" == "false" ]]; then
```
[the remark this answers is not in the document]
Deezzir: Addressed in e6335a85d00a9a5ce89a9a6947ad66079b6401b4

samuelallan72 (approved):

.github/workflows/terraform-apply.yaml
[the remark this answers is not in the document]
gabrielcocenza: It's fine for me

gabrielcocenza (dismissed):

Pjack (approved):
I like the idea and the description is clear.

H-M-Quang-Ngo:
It makes life much easier :100:  
Just a small thought, whether it is better to let the python script itself fully verify the logic and arguments to use, rather than in workflow file? So the workflow only needs to provide raw data as environment variables like:
```
- name: Update PR description with commit info
  working-directory: ./scripts
  env:
    GITHUB_TOKEN: ...
    PR_URL: ...
    PR_CREATED: ...
    COMMIT_SHA: ...
  run: |
    python3 update-pr-desc.py
```
So later if there are changes in logic or need to add more arguments to the script we don't have to touch the workflow file.

jneo8 (approved):

PR #197, opened 2025-04-03
Disable charm promote on projects to avoid regressions

gabrielcocenza:
As explained in https://github.com/canonical/charmcraft/issues/2243 and https://github.com/canonical/charming-actions/issues/157 there is a high risk of end up releasing old versions of the charm by using the workflow because of how charmcraft v2 created charms compatible with one or more bases.


samuelallan72 (approved):
LGTM

Deezzir (approved):

PR #207, opened 2025-04-14
Add a step to rename coverage reports in check workflow

H-M-Quang-Ngo:
The current coverage files generated after running unit tests or functional tests have the name `.coverage-unit` or `.coverage-func`. When downloading them to the same place, multiple unit test coverage files can continuously overwrite each other by default and reduce to a single file. The same happens with functional coverage files.  
Therefore the `coverage combine` run from `tics analysis` job in check workflow doesn't correctly achieve its purpose. For example in this [case](https://github.com/canonical/prometheus-hardware-exporter/actions/runs/14417247742/job/40435345164#:~:text=Merging%20coverage%20files), 2 unit test coverage files are downloaded, but only one file retained due to the generic name `.coverage-unit`. That also may lead to incorrect result or failure in the subsequent tics analysis steps when analyzing the incomplete `.coverage` file created by `coverage combine`. 
This PR add a step to assign the coverage reports with more specific names depending on the tests they have targeted.  A demo to correct the case above is [here](https://github.com/canonical/prometheus-hardware-exporter/commit/dd81ff1bf41325eb4970bc831862870436d2e146).


terraform-plans/templates/github/charm_check.yaml.tftpl
```diff
@@ -75,13 +75,21 @@ jobs:
     - name: Determine system architecture
       run: echo "SYSTEM_ARCH=$(uname -m)" >> $GITHUB_ENV
 
+    - name: Rename Unit Test Coverage Artifact
+      run: |
+        if [ -e ".coverage-unit" ]; then
+          mv .coverage-unit .coverage-unit-$${{ matrix.python-version }}-$${{ env.SYSTEM_ARCH }}
+        else
+          echo "No coverage file found, skipping rename"
+        fi
```
gabrielcocenza, line 84: The file names are having dots in the middle and I think this is not a good practice. E.g: `coverage-unit-3.8-x86_64`. I think it worth changing to `-`
```suggestion
      run: |
        PYTHON_VERSION_SANITIZED=$${{ matrix.python-version }}
        PYTHON_VERSION_SANITIZED=$${PYTHON_VERSION_SANITIZED//./-}
        if [ -e ".coverage-unit" ]; then
          mv .coverage-unit .coverage-unit-$${{ PYTHON_VERSION_SANITIZED }}-$${{ env.SYSTEM_ARCH }}
        else
          echo "No coverage file found, skipping rename"
        fi      
```

terraform-plans/templates/github/charm_check.yaml.tftpl
```diff
@@ -190,7 +198,15 @@ jobs:
...
           TEST_JUJU_CHANNEL: $${{ matrix.juju-channel }}
-
+      
+      - name: Rename Functional Test Coverage Artifact
+        run: |
+          if [ -e ".coverage-func" ]; then
+            mv .coverage-func .coverage-func-$${{ env.TEST_CMD_ID }}-$${{ env.SYSTEM_ARCH }}
+          else
+            echo "No coverage file found, skipping rename"
+          fi
```
gabrielcocenza, line 208: I think this block is misplaced and should be after the `Generate Safe Test Command Identifier by removing spaces and special characters` that generates the `TEST_CMD_ID`, otherwise this will fail

gabrielcocenza (changes requested):
Thanks for this.

I have just minor requests, but it's looking good

terraform-plans/templates/github/charm_check.yaml.tftpl
H-M-Quang-Ngo: That's right, my bad :sweat_smile: Correction in [737a16f](https://github.com/canonical/solutions-engineering-automation/pull/207/commits/737a16fcf3b66b634b74e0cd1337c15def6538b2)

terraform-plans/templates/github/charm_check.yaml.tftpl
H-M-Quang-Ngo: Thank you @gabrielcocenza. You're right, I made changes here [737a16f](https://github.com/canonical/solutions-engineering-automation/pull/207/commits/737a16fcf3b66b634b74e0cd1337c15def6538b2)

gabrielcocenza (approved):
LGTM.

Suggestion: Please enable the `Trim Trailing Whitspace` from your IDE so there are no "empty" spaces on our files.

PR #215, opened 2025-04-24
Move the coverage report --show-missing introduced in #207

gabrielcocenza:
We cannot run this command because empty reports of functional tests will exit with 1. After the check that there is actual data to report, we can print the report without issue and this can help for debugging.

### DIFF
--- a/terraform-plans/templates/github/charm_check.yaml.tftpl
+++ b/terraform-plans/templates/github/charm_check.yaml.tftpl
@@ -304,10 +304,10 @@ jobs:
           if [ -e "$${coverage_files[0]}" ]; then
             echo "Merging coverage files: $${coverage_files[*]}"
             coverage combine "$${coverage_files[@]}"
-            coverage report --show-missing
 
             # Check if there is actual data to report before generating XML with merged reports
             if coverage report > /dev/null 2>&1; then
+              coverage report --show-missing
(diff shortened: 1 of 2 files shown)


Deezzir (dismissed):

terraform-plans/templates/github/charm_check.yaml.tftpl
```diff
@@ -304,7 +304,6 @@ jobs:
           if [ -e "$${coverage_files[0]}" ]; then
             echo "Merging coverage files: $${coverage_files[*]}"
             coverage combine "$${coverage_files[@]}"
-            coverage report --show-missing
 
             # Check if there is actual data to report before generating XML with merged reports
             if coverage report > /dev/null 2>&1; then
```
H-M-Quang-Ngo, line 309: Not a block, but should we move it under this check for better debugging purpose 🤔 I think it's quite useful if we can have a preview of the merged report here

Pjack (dismissed):

terraform-plans/templates/github/charm_check.yaml.tftpl
gabrielcocenza: Fair enough, I've updated the PR and changed a little bit the description

H-M-Quang-Ngo (approved):

Pjack (approved):

PR #235, opened 2025-06-10
Change to use MM notification on release failure

gabrielcocenza:
GitHub bot doesn't have enough permissions to tag the team and create a notification. Using the MM bot we can notify on our channel when a release fails.

The `MATTERMOST_WEBHOOK_URL` was added on all our repos so the projects are ready to start notifying.

A small POC was done on https://github.com/canonical/charm-nrpe/pull/241 and the notification are working as expected.

image


H-M-Quang-Ngo (approved):
That's really nice :+1:

chanchiwai-ray (approved):

PR #252, opened 2025-11-06
Create dedicate template for dcgm snap

Pjack:
We need to release multiple snaps into different tracks for DCGM snap, which is not supported by the original template. It’s possible to add new configuration for DCGM, but it would be too specific and not reusable for other projects. Therefore, I propose creating a separate template specifically for DCGM.

Please check the differences between the DCGM template and the default one. This will make the review easier and may provide better solution

https://github.com/canonical/dcgm-snap/pull/73/files


Deezzir (approved):

H-M-Quang-Ngo (approved):
I think this is the best way we have, but hope there won't be other exceptions in the future. Otherwise the meaning of centrally managed will not be complete.

And later if we want to update a workflow file in `/terraform-plans/templates/github/` to apply to multiple repositories, we need to check and update the corresponding workflow file for DCGM snap as well. I think this is worth noting somewhere.

